PR #2, opened 2015-11-19
Encrypted variables were not successfully loaded to travis. Trying ne…

jykae:
…w encrypted file. Changed decrypting as first operation to travis.yml.


brylie:
We have asked not to store secrets with the source code. Please find a different way to give Travice the secrets.

PR #47, opened 2016-11-22
Dashboard refactor

brylie:
Moving work-in-progress from Apinf to Apika repository.

# Changes
- restructure/simplify dashboard markup
- reorganize dashboard logic
  - simplify logic where possible
- move dashboard filter state to URL parameters
  -  start and end date
  - granularity
  - backend
- remove hard coded date formatting
- use Bootstrap native classes rather than custom classes/styles
- refactor state handling for proxy api backends check
- remove bootstrap select logic
- refactor filter method to work with single proxy frontent prefix
- remove imperative 'loading' state UI management

# Known issues
- Loading spinners have been disabled and need to be re-added (enhancement?) (#50)
- API Select does not trigger analytics reload
- Dashboard 'no proxy added' text is behaving oddly (#51)


brylie:
Ping @bajiat @jykae @marla-singer

marla-singer:
@brylie Which behaviour is expected in the case: user has the apis with no one proxy backend configuration? Now the picture is:

dashboard/client/charts/charts.js
```diff
@@ -398,19 +393,28 @@ Template.dashboardCharts.onRendered(function () {
   $('#tick-hour').addClass('active');
```
marla-singer: Change id "tick-hour" to "hour-granularity"

dashboard/client/charts/charts.js
```diff
@@ -431,84 +432,30 @@ Template.dashboardCharts.onRendered(function () {
...
 });
 
 Template.dashboardCharts.events({
-  'change #api-frontend-prefix-form': (event) => {
+  'change #frontend-prefix': (event, templateInstance) => {
     // Prevent default form submit
     event.preventDefault();
 
     // Get reference to template instance
     const instance = Template.instance();
```
marla-singer: This variable is useless

marla-singer: This variable is useless

dashboard/client/filtering/granularity/granularity.html
```diff
@@ -0,0 +1,23 @@
+<template name="granularity">
+  <div class="btn-group" data-toggle="buttons" id="date-granularity-selector">
+    <label class="btn btn-default" for="month-input" id="month-granularity">
+      <input type="radio" name="granularity" value="month" checked>
+      {{_ "tickButtons_month" }}
```
marla-singer: Rename this i18n tag and below tag like a combination of the template name and the page element

dashboard/client/filtering/granularity/granularity.html
```diff
@@ -0,0 +1,23 @@
...
+    </label>
+
+    <label class="btn btn-default" for="week-input" id="week-granularity">
+      <input type="radio" name="granularity" value="week">
+      {{_ "tickButtons_week" }}
```
marla-singer: Rename this i18n tag as a combination of the template name and the page element

dashboard/client/filtering/granularity/granularity.html
```diff
@@ -0,0 +1,23 @@
...
+    </label>
+
+    <label class="btn btn-default" for="day-input" id="day-granularity">
+      <input type="radio" name="granularity" value="day">
+      {{_ "tickButtons_day" }}
```
marla-singer: Rename this i18n tag as a combination of the template name and the page element

dashboard/client/filtering/granularity/granularity.html
```diff
@@ -0,0 +1,23 @@
...
+    </label>
+
+    <label class="btn btn-default" for="hour-input" id="hour-granularity">
+      <input type="radio" name="granularity" value="hour">
+      {{_ "tickButtons_hour" }}
```
marla-singer: Rename this i18n tag as a combination of the template name and the page element

dashboard/client/no_apis.html
```diff
@@ -0,0 +1,12 @@
+<template name="dashboardNoApis">
+  <div class="alert alert-warning alert-text" role="alert">
+    <h3>
+      {{_ "dashboard_noApisTitle" }}
```
marla-singer: Rename this i18n tag as a combination of the template name and the page element

dashboard/client/no_apis.html
```diff
@@ -0,0 +1,12 @@
...
+    </h3>
+    <a
+      href="{{ pathFor route='addApi' }}"
+      class="btn btn-sm btn-success">
+      {{_ 'dashboard_addApiButton_text' }}
```
marla-singer: Rename this i18n tag as a combination of the template name and the page element

dashboard/client/filtering/api_select/api_select.html
```diff
@@ -1,44 +1,25 @@
...
-            {{> viewApiStatus api=api width="0.5" }}
-            &nbsp;
-            {{ api.name }}
-          </option>
-        {{/ each }}
-      </optgroup>
-    {{/ if }}
 
-    <!-- other APIs -->
-    {{# if otherApis }}
```
marla-singer: Did we cancel of idea to show other apis for admin case?

marla-singer:
@brylie Found somecase:
If you select an api and refresh the page then query parameters will have **backend** parameter. 
Is that good? Should we clear all query parameters onCreated hook?

brylie:
URL parameters should persist across page refresh. E.g. using URL parameters means that we can bookmark and share specific urls, so the dashboard populates with specific data. Clearing the URL parameters onCreated would make it so people couldn't use specific dashboard query urls.

marla-singer:
Okay, I see your point

brylie:
> Which behaviour is expected in the case: user has the apis with no one proxy backend configuration? 

The Dashboard should show text instructing user to add Proxy to an API. I am not sure that I can resolve this regression right away. Can we move it as a follow up task?

dashboard/client/no_apis.html
brylie: Done.

dashboard/client/no_apis.html
brylie: Done.

marla-singer:
> Can we move it as a follow up task?

Yep

marla-singer:
Merge it

PR #54, opened 2016-11-25
Feature/elastic search query

marla-singer:
Closes https://github.com/apinf/platform/issues/1872

### Changes
- Remove Proxy Admin API for admin case because of multiple instances of api umbrella
- Move elastic search methods to own folder and add parameter for proxy id
- Fix the bug "API Select does not trigger analytics reload"


jykae:
Reviewing

jykae:
@marla-singer Gets data but showing "no data found" layer:
nayttokuva 2016-11-28 kello 12 15 10

marla-singer:
@jykae Yes, it's old and known bug and it has an own issue: https://github.com/apinf/platform/issues/1904
But it was closed for some reason

jykae:
Notes about date picker:
- does not validate range, "To" date can be before "From" date 
- on page refresh date parameters disappear
nayttokuva 2016-11-28 kello 12 32 46

marla-singer:
@jykae Agree, Date picker must be improved. I'll create an issue for it

jykae:
@marla-singer Is there bug for this "API Select does not trigger analytics reload" that you are fixing in this PR? If then point to the issue using "Closes".

However the changes you propose look good, and solve the Definition of Done :)
So I think we can merge this and fix found bugs on separate issues.

marla-singer:
@jykae  "API Select does not trigger analytics reload" - the chart data was updated when you  were selecting different API
Example on nightly:
**Selecte "Proxy Admin API"**

joxi_screenshot_1480330452919

**Selecte another one API**

joxi_screenshot_1480330452919 2

jykae:
@marla-singer Opened found bugs, merging this.

PR #49, opened 2016-11-22
Feature/add organization view

frenchbread:
Closes https://github.com/Digipalvelutehdas/APIKA/issues/37

PR #59, opened 2016-11-28
Feature/upload organization logo

### LINKED ISSUE #43: Add organization logo
Organizations should be able to add their logo. The logo will be displayed in Organization page and in Organization catalog.

# Definition of done
* [ ] Button for adding a logo exists in Add organization view.
* [x] There is a collection for storing the logo.
* [x] Logo is displayed on single organization page.

# Wireframe
In Add Organization form
addui

After uploading a logo
addui2

In card view (if implemented)
organizationcard

In organization view, the icon would appear like this:
organizationprofile3

jykae:
Closes #43


jykae:
Waiting add organization view to be merged #49 to include upload logo.

organizations/logo/upload/client/resumable.js
```diff
@@ -0,0 +1,48 @@
...
+
+        // Update logo id field
+        Organizations.update(organization._id, { $set: { organizationLogoFileId } });
+
+        // Get success message translation
+        const message = TAPi18n.__('organizationLogo_resumable_successfully_uploaded');
+
+        sAlert.success(message);
+
+        return OrganizationLogo.resumable.upload();
```
marla-singer: It's very bad solution. You must update Organizations collection and show a message on fileSuccess event.
https://github.com/apinf/platform/blob/develop/branding/cover_photo/client/upload/resumable.js#L13-L28
In your case you add the id of non-existent image in Organization collection, tell to user that his image was updated and after then you upload the image.
I had the problems with this schema when network speed is slow. Don't do so :point_up:

jykae:
#49 got merged, continuing..

jykae:
Blocked, see comment https://github.com/Digipalvelutehdas/APIKA/issues/43#issuecomment-263560563

brylie:
This should no longer be blocked.

/cc @bajiat

jykae:
Continuing..

jykae:
Looks now like this, hover shows camera icon, click opens file browse
When file is added, previous logo is removed & updated to use new uploaded logo.

Is circle style ok or should it be square ?

nayttokuva 2016-12-05 kello 16 32 37

brylie:
> Is circle style ok or should it be square ?

We should not crop or block parts of the logo, as they are particular to organizations' branding. So, leave them at their original shape for now. We can come back to the decision at a later point.

jykae:
@brylie Discussed quickly with @bajiat and switched back to square.

brylie:
@jykae we should not cut or change the image aspect ratio. If possible, leave the image as it is, only resizing if necessary.

jykae:
thanks @brylie for vertical align rule :)

@marla-singer ready for review

marla-singer:
@jykae Cool, it works :) 
joxi_screenshot_1480952211813

Will logo changing be available in another task?

jykae:
@marla-singer If you upload new image it removes previous logo, replaces with new upload. Is that what you mean logo changing?

marla-singer:
@jykae If i want to delete current logo (in this case picture with cat) and upload another one.

marla-singer:
@jykae I found some strange behavior. I'm testing and can't undesrtand when it happens

marla-singer:
### Steps
1.Create a new organization
2. Upload logo
3. Pay attention that after logo uploading the url has **#** in the end (I used Chrome)
4. Click on the logo to change it.

### Found
File browser doesn't open. But

organizations/logo/view/client/view.js
```diff
@@ -0,0 +1,40 @@
...
+      if (organizationLogoFile) {
+        // Get API logo file URL
+        return `${Meteor.absoluteUrl().slice(0, -1) + OrganizationLogo.baseURL}/md5/${organizationLogoFile.md5}`;
+      }
+    }
+  },
+  organizationLogoExists () {
+    const organization = Template.currentData().organization;
+
+    if (organization && organization.organizationLogoFileId) {
```
marla-singer: You can simplify this one as
```javascript
return organization && organization.organizationLogoFileId
```

marla-singer:
@jykae Also it's needed to setting permission for uploading. Now non-owner can change organization logo

jykae:
@marla-singer requested fixes done. Could not find way to get organization in OrganizationLogo collection permissions and can't possibly restrict to just admin, so added permission check to "fileAdded" event. Hope thats ok for now.

marla-singer:
@jykae Your solution for permission works good as well :+1:

organizations/collection/helpers.js
```diff
@@ -1,12 +1,25 @@
...
+    // Check that user is logged in
+    if (userId) {
+      // Check if user is manager of this organization
+      const userIsManager = _.includes(this.managerIds, userId);
+
+      // Check if user is administrator
+      const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
+
+      // if user is manager or administrator, they can edit
+      return userIsManager || userIsAdmin;
```
marla-singer: And here you can check
```javascript
return this.createdBy === userId || userIsAdmin;
```

jykae:
@marla-singer I would not want restrict organization managers to just one, if customer requirements are not clear & change. That way we can be more flexible and not need schema migrations. Imho usual decision in schema modeling. Organization manager is not necessarily the one who created organization.

marla-singer:
@jykae Right, I see your point. So let's merge it

organizations/logo/view/client/view.html
```diff
@@ -1,19 +1,16 @@
 <template name="viewOrganizationLogo">
+  <div class="profile-img-container organization-file-browse">
```
brylie: Both of these classes look like they could be IDs. Consider removing one of them, or at least making one an ID for the div.

brylie: E.g. the `profile-img-container` seems redundant, while the `organization-file-browser` seems like a good ID.

organizations/client/view/header/header.less
```diff
@@ -27,31 +27,28 @@
     position: relative;
     display: inline-block;
     overflow: hidden;
+    cursor: pointer;
 }
 
 .profile-img-container img {
```
brylie: Try to remove CSS/LESS like this when possible. Rely instead on Bootstrap and semantic class names. Fewer lines of code is better.

organizations/logo/view/client/view.html
brylie: > The .img-responsive class applies max-width: 100%, height: auto, and display:block to the image.

organizations/client/view/header/header.less
jykae: @brylie I don't know way to do this layer style without additional rules. But as discussed I move these rules to "view.less" so they are closer to place they are used.

organizations/logo/view/client/view.html
jykae: @brylie Good point about ID.

jykae:
@brylie @marla-singer Done.

marla-singer:
@brylie Is everything okay?

brylie:
Jep, looks good. I was just wanting to reduce the number of custom classes and LESS code we maintain.

Issue #50, opened 2016-11-24
Design/implement Dashboard loading state notification for end user


brylie:
When the Dashboard is retrieving data, the end user may be unaware that anything is happening. In order for the user to understand that something is happening, we need to show some loading indicator. The previous loading indicator code was removed as part of a refactor.

frenchbread:
Interested

brylie:
Thanks for taking this task @frenchbread.

Previously, we were using imperative code wrapped within other functions, which made it difficult to keep track of what the code was doing with so many parts. Lets try to use declarative code where possible, e.g. template helpers that observe the state of `chartData`.

frenchbread:
@brylie Agreed, previous solution what a bit messy.

PR #64, opened 2016-11-29
Code refactoring of Elastic Search

### LINKED ISSUE #61: Ensure dashboard renders charts with ElasticSearch query results
Currently, the dashboard charts are not being rendered with data, when API backend has been selected from the Select API list.

# General design
screenshot_20161129_113533

The code design **might** follow this flow:

1. user selects an API from the list
2. `getProxyData` is called with the selected `backendId` and Dashboard query parameters (timeframe, granularity)
    -  on the server, the `getProxyData` method fetches the proxy document, and calls `getElasticData` with the `proxyId`
3. `getElasticData` 
    1. gets the proxy document (URL, type, etc), 
    2. constructs the Elasticsearch query object, and 
    3. makes the HTTP request, 
    4. throw an error if needed
    5. return response data to the client

# Definition of done
* [x] When selecting API backend, ElasticSearch query is made on the server to the right proxy.
* [x] The result of the query is passed to the client.
* [x] Dashboard charts update with the received data.
* [x] Dashboard charts show only data for the selected API.

Related to #50.

marla-singer:
Closes #61 
Closes https://github.com/apinf/platform/issues/1956
### Changes
- Added server method **getProxyData**:
    - fetches proxy document
    - checks of existing needed proxy parameters
    - constructs and returns needed proxy data
- Deleted two elastic search methods associated with definition URL (move this part of getProxyData method)
- Refactored getElasticSearchData method: 
    - constructs the Elastic Search query object
    - makes the HTTP request
- Move getElasticQuery to own server method (create query for apiUmbrella proxy type)


dashboard/client/dashboard.js
```diff
@@ -16,113 +17,37 @@ Template.dashboard.onCreated(function () {
   // Keeps ES data for charts
   instance.chartData = new ReactiveVar();
 
-  instance.getChartData = function (params, proxyId) {
+  instance.getChartData = function (backendId) {
     return new Promise((resolve, reject) => {
-      Meteor.call('getElasticSearchData', params, proxyId, (err, res) => {
+      Meteor.call('getElasticSearchData', backendId, (err, res) => {
```
brylie: How do we pass parameters to this method? E.g. we may have the following parameters in the Elasticsearch query:

- proxy backend
- dateFrom
- dateTo
- granularity (for aggregation)

marla-singer:
@brylie Let's review it after merging https://github.com/Digipalvelutehdas/APIKA/pull/69

brylie:
OK, cool.

marla-singer:
@brylie Ready.

elasticsearch/server/methods.js
```diff
@@ -1,18 +1,23 @@
...
-        params = Meteor.call('getElasticQuery', proxyData.frontendPrefix, analyticasData);
+        params = Meteor.call('getElasticQueryUmbrella', proxyData.frontendPrefix, analyticsData);
+        break;
+      case 'emqtt':
+        // TODO: Add fronted prefix parameter if eMQTT protocol has it
+        params = Meteor.call('getElasticQueryEmqtt');
         break;
       default:
-        throw new Meteor.Error('Unknow proxy type');
+        const message = TAPi18n.__('dashboard_errorMessage_unknownProxyType');
```
brylie: I am not sure that localzation works on the server. I.e. the 'user language' is only stored in the client, to my knowledge. So, we may need to catch the error in the client, and internationalize there.

dashboard/server/methods.js
```diff
@@ -14,28 +15,39 @@ Meteor.methods({
...
+            throw new Meteor.Error(message);
+        }
+
+        // Return proxy data
         return response;
       }
 
       // Create & throw message about error
-      const message = `No one proxy with ID ${proxyBackend.proxyId} was found`;
+      const message = TAPi18n.__('dashboard_errorMessage_noProxyFound');
```
brylie: We may need to handle localization on the client. The server code may not have awareness of the user language.

elasticsearch/server/methods.js
```diff
@@ -78,17 +83,21 @@ Meteor.methods({
...
       // Set end date (less than or equal to) for analytics timeframe
-      params.body.query.filtered.filter.range.request_at.lte = analyticasData.analyticsTo;
+      params.body.query.filtered.filter.range.request_at.lte = analyticsData.analyticsTo;
     }
 
     return params;
   },
+  // TODO: Add right constructor
+  getElasticQueryEmqtt () {
+    return { size: 1000 };
```
marla-singer: I don't know how to correctly construct the query for eMQTT.
I saw this example in @frenchbread work: https://github.com/apinf/emq-analytics-dashboard/blob/master/dashboard/client/index.js#L132
Now it's only placeholder

dashboard/server/methods.js
marla-singer: Okay

dashboard/client/dashboard.js
```diff
@@ -16,116 +16,49 @@ Template.dashboard.onCreated(function () {
...
+        analyticsTo,
+        granularity,
+      };
+
+      // Check of existing needed proxy data
+      Meteor.call('getProxyData', backendParameter, (error, result) => {
+        // if it was not error
+        if (!error) {
+          // Provide proxy data to elastic search
+          instance.getChartData(result, analyticsData)
```
brylie: From reading the code above, the `analyticsData` parameter seems to be more related to Elasticsearch query parameters, so might be better named as `queryParameters`, `filterParameters`, or something similar.

Semantically, `anyliticsData` sounds like something you would render charts with, rather than pass in to a query function (`getChartData`).

dashboard/client/dashboard.js
```diff
@@ -16,116 +16,49 @@ Template.dashboard.onCreated(function () {
   // Keeps ES data for charts
   instance.chartData = new ReactiveVar();
 
-  instance.getChartData = function (params, proxyId) {
+  instance.getChartData = function (proxyData, analyticsData) {
     return new Promise((resolve, reject) => {
-      Meteor.call('getElasticSearchData', params, proxyId, (err, res) => {
+      Meteor.call('getElasticSearchData', proxyData, analyticsData, (err, res) => {
```
marla-singer: `analyticsData` consists parameters as "dateFrom, dateTo and granularity". Do you think `queryParameters` is more obviously?

dashboard/client/dashboard.js
marla-singer: I choose `filterParameters`

dashboard/client/dashboard.js
```diff
@@ -16,116 +16,49 @@ Template.dashboard.onCreated(function () {
...
-    }
-
-    return params;
-  };
-
   instance.autorun(() => {
     if (instance.subscriptionsReady()) {
+      // The main Elastic Search Logic
       // Cleanup previous message if one exists
       $('.charts-holder>#no-chart-data-placeholder').remove();
```
marla-singer: I suggest to create another issue for that. This code part is used here too:
https://github.com/Digipalvelutehdas/APIKA/blob/develop/dashboard/client/charts/charts.js#L422-L427

elasticsearch/server/methods.js
```diff
@@ -1,49 +1,103 @@
...
+              range: {
+                request_at: {},
+              },
+            },
+          },
+        },
+        sort: [
+          {
+            request_at: {
+              order: 'desc',
```
brylie: Is there any reason that we might want to change the sort order of the `request_at` field? For now, this is probably OK, but we may want to parameterize this sometime.

elasticsearch/server/methods.js
```diff
@@ -1,49 +1,103 @@
...
+      // Set end date (less than or equal to) for analytics timeframe
+      params.body.query.filtered.filter.range.request_at.lte = analyticsData.analyticsTo;
     }
 
-    throw new Meteor.Error('Elasticsearch is not defined');
+    return params;
+  },
+  // TODO: Add right constructor
+  getElasticQueryEmqtt () {
+    return { size: 1000 };
```
brylie: What is the `size` parameter, and why is it `1000`. It might be good to at least indicate that as a comment.

brylie:
This PR is looking really good. I just have a few questions and suggestions, see above.

elasticsearch/server/methods.js
marla-singer: I don't know how to correctly construct the query for eMQTT.
I saw this example in @frenchbread work: https://github.com/apinf/emq-analytics-dashboard/blob/master/dashboard/client/index.js#L132
Now it's only placeholder

brylie: OK, cool. Perhaps add a comment like:

``js
// TODO: describe the 'size' attribute and why it is '1000'
```

/cc @frenchbread

marla-singer:
@brylie Fixed remarks

elasticsearch/server/methods.js
frenchbread: `1000` value was added for testing. Here we should provide bigger value.

> I think lucene syntax is supported so:
   `http://localhost:9200/foo/_search?pretty=true&q=*:*`
    Size defaults to 10, so you may also need &size=BIGNUMBER to get more than 10 items.(where BIGNUMBER equals a number you believe is bigger than your dataset)

Reference: http://stackoverflow.com/questions/8829468/elasticsearch-query-to-return-all-records

dashboard/client/dashboard.js
brylie: Here is the related issue:
https://github.com/Digipalvelutehdas/APIKA/issues/50

PR #65, opened 2016-11-30
Prev., Next button IE fix

nnn:
Closes https://github.com/apinf/platform/issues/1922


catalogue/client/catalogue.js
```diff
@@ -238,6 +238,8 @@ Template.catalogue.events({
   },
   // Pagination
   'click #prev-page': function (event, instance) {
+    event.preventDefault();
```
brylie: Rather than supressing the event here, just remove the [anchor tag](https://github.com/Digipalvelutehdas/APIKA/blob/develop/catalogue/client/catalogue.html#L125 from the forward and back buttons. Buttons don't need to be anchors.

Basically, and we've said this before, whenever we are using `href="#`, we probably don't need to be using an anchor tag at all. Usually a `<button>` tag will suffice.

catalogue/client/catalogue.js
```diff
@@ -246,6 +248,8 @@ Template.catalogue.events({
     }
   },
   'click #next-page': function (event, instance) {
+    event.preventDefault();
```
brylie: Rather than supressing the event here, just remove the [anchor tag](https://github.com/Digipalvelutehdas/APIKA/blob/develop/catalogue/client/catalogue.html#L125 from the forward and back buttons. Buttons don't need to be anchors.

Whenever we are using `href="#"`, we probably don't need to be using an anchor tag at all. Usually a `<button>` tag will suffice.

catalogue/client/catalogue.js
```diff
@@ -82,7 +82,7 @@ Template.catalogue.onCreated(function () {
   });
 });
 
-Template.catalogue.onRendered(function () {
+Template.catalogue.onRendered(() => {
```
nnn: Interesting. Good to know.

nnn:
Will close it for now. Will need more time for fixing navigation buttons.
Simple change to `<button>` didn't work.

Issue #55, opened 2016-11-28
"No data found" layer when data is available on Dashboard


jykae:
Found when reviewing #54  
Possibly related to https://github.com/apinf/platform/issues/1904

## Steps to reproduce
1. Add 2 APIs, other dummy API without data & other Real API with some query data
2. Switch between APIs using Select list

## Expected
- When data is available, don't show "No data found" layer

## Actual
- Showing "No data found" layer when data is available

PR #70, opened 2016-12-01
Remove noDataFound template on autorun hook

marla-singer:
Closes https://github.com/Digipalvelutehdas/APIKA/issues/55

### DIFF
--- a/dashboard/client/dashboard.js
+++ b/dashboard/client/dashboard.js
@@ -105,6 +105,9 @@ Template.dashboard.onCreated(function () {
 
   instance.autorun(() => {
     if (instance.subscriptionsReady()) {
+      // Cleanup previous message if one exists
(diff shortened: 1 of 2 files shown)


marla-singer:
@jykae Could you please review it?

dashboard/client/dashboard.js
```diff
@@ -105,6 +105,9 @@ Template.dashboard.onCreated(function () {
 
   instance.autorun(() => {
     if (instance.subscriptionsReady()) {
+      // Cleanup previous message if one exists
+      $('.charts-holder>#no-chart-data-placeholder').remove();
```
brylie: Eventually, we should not manipulate the DOM imperatively. In other words, we should rely on template helpers, charting libraries, or some declarative means to change DOM text. This includes showing and hiding placeholder text, loading icons, etc.

Issue #51, opened 2016-11-24
Fix Dashboard 'no API added' text


brylie:
The Dashboard should inform a user to add an API with Proxy settings when they have not done so already. This was working prior to the recent Dashboard refactor, but needs attention now.

saralavanip:
verified this resolved issue on https://staging.apinf.io

PR #77, opened 2016-12-01
Bugfix/no api text

marla-singer:
Closes https://github.com/Digipalvelutehdas/APIKA/issues/51

### Changes
- Add Template.subscriptionsReady to exclude oddly behaviour


marla-singer:
@brylie Could you please review it?

PR #95, opened 2016-12-05
Sync Apinf 0.36.0

brylie:
Sync small changes from Apinf release 0.36.0


brylie:
@shaliko it looks like our Dokerfile got out of sync between Apinf and Apika. Be advised, we are merging these changes, and need to sync the changes between the two projects going forward.

PR #112, opened 2016-12-08
Show organization apis

### LINKED ISSUE #74: Showing organization APIs on organization page
# User story
> As an organization manager,
I want to show my organization's APIs on the organization page
so that I can promote them.

# Definition of done
* [x] Organization page contains an API catalog that lists only APIs connected to that organization
* [x] The catalog only shows API cards, i.e. there is no option to toggle to table

# Wireframe

1. new API card design
It includes new label to indicate what lifecycle phase the API is currently in.
cardapi

2. This is how APIs appear in a organization view
In Organization profile, members within an organization (future option) as well as APIs appear as cards/blocks.
(with future implementation works) APIs can be filtered by clicking of labels (quick filters: visibility, API Lifecycle) or accessing the filter menu.
organizationprofile3
If an organization has a lot of APIs that can be seen in a single view and have to browse downwards for viewing them, we can implement infinite scrolling and option to go back on top

open question: should we show settings option in a modal or a tabview (working on the wireframes)

marla-singer:
Closes #74

Used standart catalog template


jykae:
@marla-singer Reviewing after lunch 🍴

organizations/client/view/view.html
```diff
@@ -1,5 +1,13 @@
 <template name="organizationProfile">
+  <div class="container">
     {{# if Template.subscriptionsReady }}
-    {{> organizationProfileHeader organization=organization }}
+      {{> organizationProfileHeader organization=organization }}
+      <div class="row">
+        <label class="h3">
```
jykae: @marla-singer Why using label with class="h3" here and not \<h3\> for example?

marla-singer: Tag <h3> has a semantic meaning for search engines. In this case, the word APIs do not have the necessary sense to specify the tag.
In a nutshell, because of Search Engine Optimization (SEO)

jykae:
Overall good work.

Align header with other elements if possible, and use appropriate html header element for it
nayttokuva 2016-12-09 kello 12 29 16

marla-singer:
@jykae I remade the header of organization page
joxi_screenshot_1481529162370

jykae:
@marla-singer great! just remove unused classes from "header.less" or whole file if not needed
Then I think we are ready to merge :)

marla-singer:
@jykae Done :muscle:

PR #113, opened 2016-12-08
REST API: filter APIs by organization

### LINKED ISSUE #110: Add filtering by organization
- override action() for GET all to handle filtering Apis by organization ID with query param

jykae:
Closes #110


marla-singer:
@jykae Reviewing

marla-singer:
@jykae REST works as expected :+1:

marla-singer:
What behaviour is expected if we come with another query parameter? It returns all apis (because of conditions)
Is it okay and go on? Or it will be updated on the next step?

jykae:
@marla-singer Maybe it should give error, queryparam does not exist or something?  
We don't have definitions for error cases yet, so have to probably think something smart & obvious behaviour for error cases. Do you want to me add that behaviour in this PR?

jykae:
@marla-singer I add that error case, and move api files based on discussion that we had with @brylie about architecture (band playing & football)

marla-singer:
@jykae Yep, I think 40 BadRequest it's good.

jykae:
@marla-singer ready. Moved core API under "core" folder, Apis collection related under "apis" folder, added error handling for query params.

apis/server/api.js
```diff
@@ -0,0 +1,99 @@
...
+            description: 'List of APIs.',
+          },
+          400: {
+            description: 'Bad query parameters.',
+          },
+        },
+      },
+      action () {
+        /*
+          Expected response object structure:
```
marla-singer: Is comment unnecessary,

jykae: @marla-singer ok, I can remove it, would have left it for reminder what object structure Restivus uses to keep consistent with generated routes response structure.

marla-singer:
@jykae Merge it

PR #117, opened 2016-12-12
Add id for dropdown proxy list

marla-singer:
The first point of this task https://github.com/apinf/platform/issues/1961


Nazarah:
please make the id parameter name as "proxy-selector" instead of "proxy-id-selector".
ince this is a type "id" so giving the parameter name something with "id" seems confusing
and not follows normal convention of coding

Issue #115, opened 2016-12-09
Consistency between API and Organization catalogs


brylie:
Currently, the Organization and API catalogs are distinct in several ways:
- The API catalog has different pagination implementation than the Organization catalog
- The Organization and API catalogs have different sort/filter functionality
- The Organization and API catalogs do not use URL parameters for sorting/filtering

# Task
Unify the functionality and style of the Organization and API catalog as much as possible. Make sure to share code and styles as much as possible.

# Definition of done
- [x] APIs catalog has pagination feature
- [x] Both APIs and Organizations catalogs support sorting and filtering
  - [x] Sorting and filtering parameters are stored as URL variables

marla-singer:
@brylie Implemented pagination on APIs catalog and sorting. Now I am working on filtering

bajiat:
@marla-singer Do you want this issue to be split into two different issues?

marla-singer:
@bajiat Yes,please. I'd like to separate redesing of catalog card to different issue

brylie:
@marla-singer I am revising the issue now.

brylie:
Created issue #118 as a follow-up task.

Nazarah:
@marla-singer @brylie : Have you decided on what categories you will consider for both the sorting and filtering option?

brylie:
Here are some:

Sort:
- alphabetical
- date created
- popularity

Filter:
- My APIs
- My Organization(s)

The query parameters don't need to be exact between Organizations and APIs, but should overlap somewhat.

Nazarah:
@brylie : for the organization I suggest to exclude the sort category "popularity". From the weekly update meetings @kyyberi  suggested that we don't need to include popularity/rating options for organizations.

One more suggestion is that we can include "My APIs in Organizations". This would filter the organizations to which an API owner has his/her APIs connected.

PR #119, opened 2016-12-12
Feature/apis catalog pagination

marla-singer:
Part of #115 
- Implemet pagination in APIs catalog via package
- Setting sorting & filtering
### Changes
- Remove setQueryParams from template logic (for APIs catalog)
- Add setQueryParams on routing (for APIs catalog)


users/client/lib/router.js
```diff
@@ -47,15 +47,12 @@ FlowRouter.route('/settings/profile', {
...
-  action: function () {
-    BlazeLayout.render('masterLayout', { main: 'signOut' });
-  },
+  triggersEnter: [
+    function () {
+      Meteor.logout(() => {
+        FlowRouter.go('catalogue');
+      });
+      // Page render
+      BlazeLayout.render('masterLayout', { main: undefined });
```
jykae: @marla-singer what that undefined does? is it bug? 🤔

jykae: Looks like it works, could you make better comment for it?

jykae: @marla-singer ok, could you add that "force render" explanation to comment somehow short way as it is not obvious what it does. Eg. "Use undefined to force page render when route is not changed"

jykae:
@marla-singer Overall good work. 👍

marla-singer:
@jykae Done :)

PR #120, opened 2016-12-12
Enable authentication for REST API

### LINKED ISSUE #109: REST API authentication
- enable defaultAuth
- routes for authentication, show in swagger?
- define explicitly which routes require authentication

jykae:
Closes #109 

## For reviewer
- check out documentation for auth usage https://github.com/kahmali/meteor-restivus#default-authentication
- try out as anonymous user (error) & authenticated user (login, make request, logout)

### DIFF
--- a/apis/server/api.js
+++ b/apis/server/api.js
@@ -6,7 +6,7 @@ import _ from 'lodash';
 // /rest-api/v1/apis/:id for Apis collection
 ApiV1.addCollection(Apis, {
   excludedEndpoints: ['post', 'put', 'delete'],
-  routeOptions: { authRequired: false },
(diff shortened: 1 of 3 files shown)


jykae:
Tokens generated by Restivus don't have "when" property. I added proposal to add it to enable token expiration. https://github.com/kahmali/meteor-restivus/issues/252

bajiat:
@brylie Could you review this?

jykae:
@brylie Authentication tokens live forever if not revoked, and user can create as much as wanted. Expiration might be also bad idea, in case of APIs and systems interoperability. So I think it can be ignored at this point.

PR #123, opened 2016-12-12
Feature/loading state

### LINKED ISSUE #50: Design/implement Dashboard loading state notification for end user

frenchbread:
Closes #50 

# Changes
- Modified i18n tag
- Added new i18n tag
- Added logic to `dashboardCharts` template in `dashboard/client/charts/charts.html`
- Added `instance.loadingState()` reactive variable that keeps loading state value
- Added `instance.updateLoadingState(ChartData)` function that takes care of updating loading state
- Cleanups


frenchbread:
@Digipalvelutehdas/apika-tiimi Please review.

bajiat:
@marla-singer Could you review this PR?

marla-singer:
@bajiat Yes, after lunch

marla-singer:
@frenchbread Reviewing

dashboard/client/charts/charts.js
```diff
@@ -403,6 +403,32 @@ Template.dashboardCharts.onRendered(function () {
 
     const frontendPrefix = instance.frontendPrefix.get();
 
+    // Init elements on dashboard
```
brylie: This is still very imperative. Is there a way to use template helpers, for example?

This file is over 400 lines long, partially because it tries to do so much, imperatively.

marla-singer:
@frenchbread Works as expected. What about @brylie comment as for me it must be special issue for refactoring code of chart.js. But if you can do it via helpers quickly I won't merge it and will wait implementation

brylie:
It might be possible to do this without even needing a template helper. I.e. `chartData` seems to be available as a reactive data source directly within the template:

```html
{{# unless chartData }}
  <div id="no-chart-data-placeholder">
    {{_ "dashboardCharts_placeholder_loadingData" }}
  </div
{{/ unless }}
```

brylie:
In concrete terms, using `domElement.append("<some-markup>... </some-markup>")` is an anti-pattern in our project. We more commonly use Blaze/Handlebars block helpers to reactively toggle blocks of markup, such as the dashboard charts placeholder.

Is there a solution we can find that is closer to the suggestion above?

brylie:
Breaking things down, it seems like we want to track two different states:
- loading data
- data not found

Right now, there is a nested loop handling both of those, within an autorun function.

I am just trying to see if we can decompose this problem any better, and move the template markup to the template.

brylie:
Go ahead and merge, if you feel there is not a suitable solution that could lessen the lines of code and complexity within charts.js.

frenchbread:
@marla-singer @brylie Thanks for review. Good catch about html markup in JS code. I'll refactor it to use template helper & reactive variable.

core/lib/i18n/en.i18n.json
```diff
@@ -197,7 +197,8 @@
   "dashboardDataTable_timeStamp": "Date",
   "dashboardDataTable_title": "Analytics data",
   "dashboardDataTable_pageCount": "Page",
-  "dashboardCharts_placeholder_noDataFound": "No data found. <br/> Try changing filtering options to get some analytics data.",
+  "dashboardCharts_placeholder_noDataFound": "No data found. Try changing filtering options to get some analytics data.",
+  "dashboardCharts_placeholder_loadingData": "Fetching latest data..",
```
nnn: Typo here, should be "..."

dashboard/client/charts/charts.js
```diff
@@ -19,6 +19,14 @@ Template.dashboardCharts.onCreated(function () {
...
   instance.frontendPrefix = new ReactiveVar();
 
+  // Variable that keeps loading state value
+  // States:
+  // 0 - not loading (by default)
+  // 1 - loading in progress
+  // 2 - loaded && data NOT found
+  // 3 - loaded && data found
+  instance.loadingState = new ReactiveVar(0);
+
```
marla-singer: @frenchbread Why do you not use semantic word instead of numbers? 
E.g.
```javascript 
instance.loadingState = new ReactiveVar('no-loading');
...
if (!chartData) {
      instance.loadingState.set('loading');
     } else {
      // Check if chart data was found
       if (chartData.length > 0) {
         // If found, Hide all messages
         instance.loadingState.set('data-found');
       } else {
         // If not, display "not found" message to user
         instance.loadingState.set('data-no-found');
       }
     }
   };
```

And compare string status

frenchbread: @marla-singer Good point. Using phrases as reference to "state" is more complex I think since it's easy to make typos there.

core/lib/i18n/en.i18n.json
frenchbread: Fixed

dashboard/client/charts/charts.js
frenchbread: But I like the idea actually. Do you think I should change it to use 'phrases` instead of numbers?

marla-singer: @frenchbread Yes, if it isn't pretty complex, it will be great to implement via phrases.
I think helper `isEqual` doesn't need any changes

marla-singer:
@frenchbread I deleted label "ready for review". You can pick it again when PR will be ready

frenchbread:
Please review 👀 . I've change int values to str. Looks more readable this way, and no "translation map" needed anymore. 😃

marla-singer:
@frenchbread Good. Merge it

PR #137, opened 2016-12-14
Sorting as query params for organization catalog

marla-singer:
Part of #115 
- Add and configure query parameters for sorting & filtering (as future feature) in organization catalog


bajiat:
@nnn Could you review this PR?

PR #166, opened 2016-12-17
Required proxy fields

### LINKED ISSUE #104: A new proxy can be added without filling up mandatory information for URL and Elastic Search
# Reproduction step
1. sign in as APINF admin.
2. Go to profile->Proxy
3. Click Add proxy
4. 1st select proxy type without filling other fields. (mandatory fields don't get populate after selecting proxy type)
5. Fill up name and description and save.

# Result:
Proxy can be added without filling mandatory fields (elastic search and URL).
This behavior doesn't occur if fields are filled serially.
proxy

# Expected Result:
no matter what order the fields are filled up, on selecting a proxy from the dropmenu, the mandatory fields should appear in the instant.

marla-singer:
Closes #104
Form saves result without proxy data both apiUmbrella and emqtt type
- Create validation function 
- Checking for validation on before-insert autoform hook
- Delete useless **optional** field in schema (on default _optional: false_)

### DIFF
--- a/core/lib/i18n/en.i18n.json
+++ b/core/lib/i18n/en.i18n.json
@@ -385,6 +385,8 @@
   "proxyBackendForm_informText_noOneSelectedProxy": "No one proxy was selected.",
   "proxyForm_header_text": "Add Proxy",
   "proxyForm_saveButton_text": "Save",
+  "proxyForm_successText": "Proxy settings were saved",
+  "proxyForm_errorText": "All fields with asterisk symbol are required",
   "proxyItem_editProxy": "Edit",
   "proxyItem_removeProxy": "Remove",
   "proxyItem_type": "Type",
--- a/proxies/client/form/autoform.js
+++ b/proxies/client/form/autoform.js
@@ -1,14 +1,45 @@
 import { Meteor } from 'meteor/meteor';
 import { AutoForm } from 'meteor/aldeed:autoform';
 import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
+import { TAPi18n } from 'meteor/tap:i18n';
+import { sAlert } from 'meteor/juliancwirko:s-alert';
+
+import validateSchema from './validation';
 
 AutoForm.addHooks('proxyForm', {
+  before: {
+    insert (formData) {
+      if (formData.name && formData.description && formData.type) {
+        // Get type of proxy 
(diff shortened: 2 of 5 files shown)


brylie:
While this solution is clever and elegant, there may be a [solution that is more aligned with simple schema](https://github.com/aldeed/meteor-simple-schema#make-a-field-conditionally-required). In that approach, the `apiUmbrella` or `emqtt` sub-fields would look at the `proxyType` field and perform validation based on the proxy type selected. This will get [much easier in SimpleSchema v2](https://github.com/aldeed/meteor-simple-schema/issues/618), so lets keep this work relatively close to how Simple Schema is designed to work.

brylie:
Basically, what we are trying to do is **define validation in the schema**.

marla-singer:
@brylie Of course it was my first decision to make validation with simple schema. But we have three independent schemas. `apiUmbrella` and `emqtt` schemas don't know anything about `proxy` schema. We can't use[ this solution](https://github.com/aldeed/meteor-simple-schema#make-a-field-conditionally-required) because we don't know in which moment we must make a field required.

brylie:
Even with multiple sub-schemas, the `proxyType` should be available at the time of validation. I.e. when a document is validated, Simple Schema will be using the combined schema -- despite us breaking it into parts for extensibility.

brylie:
I am willing to merge this solution, if there is no way we can validate closer to the schema. So, lets leave this branch/PR open.

marla-singer:
> Try a quick test to validate or refute the idea, e.g.:
// Inside of custom validation function for emqtt URL field
console.log(this.field('proxyType'));

@brylie Nothing useful
```javascript 
{isSet: false, value: undefined, operator: undefined}
```

brylie:
Ah, sorry. It should have been:

```js
// Inside of custom validation function for emqtt URL field
console.log(this.field('proxyType')).value;
```

marla-singer:
> Did you set a value for the Proxy Type field, via the form?
Sure

brylie:
OK, I am thinking that our schemas don't need to be so complicated for right now. I.e. we are defining them in separate files for modularity, but they are hard coded inside of index.js anyway (in terms of imports and listed within `_.assign()`. Lets move all of the schemas back together inside of `index.js`, that way we can use the custom validation function.

PR #185, opened 2016-12-22
Remove subscription need access organization profile

### LINKED ISSUE #169: Anonymous user can't see any Organization profile
Anonymous user can't see any Organization profile, looks like the empty page.
### Steps
1. Go to nightly.apinf.io
2. Sign out if you was sign in
3. Navigate to Organizations Catalog
4. Select any organizations

### Found
Empty page without information about organization

### Expected
Any user can see organization profile

manzapanza:
Closes #169 

Now anonymous users can access to the organization profile

### DIFF
--- a/apis/collection/server/publications.js
+++ b/apis/collection/server/publications.js
@@ -1,15 +1,14 @@
 import { Meteor } from 'meteor/meteor';
 import { Apis } from '../';
 
-Meteor.publish('allApiBackends', function () {
+Meteor.publish('allApiBackends', function (userId) {
   // Check if the user is signed in
   if (this.userId) {
     // Return all API Backends
     return Apis.find();
   }
-
-  // Return nothing
-  return null;
+  // Return only public apis
+  return Apis.find({ isPublic: true });
 });
 
 Meteor.publish('myManagedApis', function () {
--- a/organizations/client/view/view.html
+++ b/organizations/client/view/view.html
@@ -1,17 +1,15 @@
 <template name="organizationProfile">
-    {{# if Template.subscriptionsReady }}
-      {{> organizationProfileHeader organization=organization }}
-    <div class="container">
-      <!--container-fluid addes padding-left -->
-      <div class="container-fluid row">
-        <label class="h3">
-          {{_ 'organizationProfile_bigText_apisCount'  }}
-        </label>
-        <label>
-          {{_ 'organizationProfile_smallText_apisCount' managedApisCount}}
-        </label>
-      </div>
-      {{> catalogueGrid apis=managedApis }}
+  {{> organizationProfileHeader organization=organization }}
+  <div class="container">
+    <!--container-fluid addes padding-left -->
+    <div class="container-fluid row">
+      <label class="h3">
+        {{_ 'organizationProfile_bigText_apisCount'  }}
+      </label>
+      <label>
+        {{_ 'organizationProfile_smallText_apisCount' managedApisCount}}
+      </label>
     </div>
-    {{/if}}
+    {{> catalogueGrid apis=managedApis }}
+  </div>
 </template>
--- a/organizations/client/view/view.html
+++ b/organizations/client/view/view.html
@@ -1,15 +1,18 @@
 <template name="organizationProfile">
-  {{> organizationProfileHeader organization=organization }}
-  <div class="container">
-    <!--container-fluid addes padding-left -->
-    <div class="container-fluid row">
-      <label class="h3">
-        {{_ 'organizationProfile_bigText_apisCount'  }}
-      </label>
-      <label>
-        {{_ 'organizationProfile_smallText_apisCount' managedApisCount}}
-      </label>
+  {{# if Template.subscriptionsReady }}
+    {{> organizationProfileHeader organization=organization }}
+    <div class="container">
+      <div class="row">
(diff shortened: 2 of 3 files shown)


brylie:
It seems like we should be able to wait for the subscription to be ready. What subscriptions are active on the page? Is one of the subscriptions 'broken' or unneeded?

brylie:
E.g. check the purpose and function of [/organizations/client/view/view.js:L17](https://github.com/Digipalvelutehdas/APIKA/blob/develop/organizations/client/view/view.js#L16-L17)

brylie:
The `allApiBackends` subscription is probably the issue here. There are a few things that should be changed:

# Error
- there is an error in the `allApiBackends` subscription, where the [allApiBackends publication](https://github.com/Digipalvelutehdas/APIKA/blob/develop/apis/collection/server/publications.js#L4) is not designed to take an argument

# Fix
We don't need to subscribe to `allApiBackends` on the organization page, since the organization only owns a subset of API Backends. So,

For now, remove the subscription. We will add the subscription back once we have built the relationship between APIs and Organizations.

brylie:
Ping @marla-singer

brylie:
There is also some markup that needs revising. I will make notes inline in the code review.

organizations/client/view/view.html
```diff
@@ -1,17 +1,15 @@
...
-        </label>
-        <label>
-          {{_ 'organizationProfile_smallText_apisCount' managedApisCount}}
-        </label>
-      </div>
-      {{> catalogueGrid apis=managedApis }}
+  {{> organizationProfileHeader organization=organization }}
+  <div class="container">
+    <!--container-fluid addes padding-left -->
+    <div class="container-fluid row">
```
brylie: Remove this div. One container is sufficient for this template.

organizations/client/view/view.html
```diff
@@ -1,17 +1,15 @@
...
+      <label class="h3">
```
brylie: Change this to an `<h3>` tag with no class.

organizations/client/view/view.html
```diff
@@ -1,17 +1,15 @@
...
+        {{_ 'organizationProfile_bigText_apisCount'  }}
+      </label>
+      <label>
```
brylie: Move this inside of the `<h3>` tag. Change this tag from `<label>` to `<small>`.

organizations/client/view/view.html
```diff
@@ -1,17 +1,15 @@
...
+        {{_ 'organizationProfile_smallText_apisCount' managedApisCount}}
+      </label>
     </div>
-    {{/if}}
+    {{> catalogueGrid apis=managedApis }}
```
brylie: Move this inside of the `Template.subscriptionsReady` block.

brylie:
We can probably remove the `allApiBackends` publication.

- there aren't many, or possibly any, situations where we want a user to have access to "All API Backends"
  - we do want users to access "All Public APIs" and "All APIs where the user is a manager"

marla-singer:
> there aren't many, or possibly any, situations where we want a user to have access to "All API Backends"

@brylie Without subsription the APIs view part doesn't work on Organization profile

brylie:
> Without subsription the APIs view part doesn't work on Organization profile.

A 'proper' solution here would be to build a publication specifically for APIs related to an organization:

```js
Meteor.publish('organizationApis', function () {
  // Return a cursor of APIs related to the Organization
  // Make sure only public APIs are returned if current user is not Organization manager
});
```

apis/collection/server/publications.js
```diff
@@ -1,15 +1,14 @@
 import { Meteor } from 'meteor/meteor';
 import { Apis } from '../';
 
-Meteor.publish('allApiBackends', function () {
+Meteor.publish('allApiBackends', function (userId) {
```
brylie: The `userId` argument is not necessary.

manzapanza: Ok, I'll fix it

apis/collection/server/publications.js
```diff
@@ -1,15 +1,14 @@
...
   // Check if the user is signed in
   if (this.userId) {
     // Return all API Backends
     return Apis.find();
   }
-
-  // Return nothing
-  return null;
+  // Return only public apis
+  return Apis.find({ isPublic: true });
```
brylie: We can remove it, since it is not used.

Subscribing to `allApis` on the organization page is over-subscription, meaning that it is too much data and not relevant to the feature. The Organization APIs feature needs a **specific publication**, called `organizationApis`, which returns only APIs linked to the organizaiton.

manzapanza: Ok!

apis/collection/server/publications.js
brylie: @jykae, can you assist here? We need an `organizationApis` publication.

apis/collection/server/publications.js
jykae: @manzapanza I could create PR to add that publication, forgot that part, I let you know, when it's on place and you can pull it here from develop.

jykae: PR #193 adds publication to organizationApis, asked @brylie to review

brylie:
Closing, as fixed in different branch.

PR #199, opened 2017-01-02
Feature/rest api route

### LINKED ISSUE #162: Unify organizations and apis routes
Follow the recommendation from #86 and change the routes for 
- organization catalog
- individual organization profile
- API catalog
- individual API profile
- REST api

Also, consider the impact of the change.

brylie:
Closes #162 

# Changes
- Change api route to `/api`
- Remove hardcoded values from swagger definition
  - Apika specific language
  - Broken link to terms of service
- Make swagger fields more generic

### DIFF
--- a/core/server/api.js
+++ b/core/server/api.js
@@ -2,7 +2,7 @@ import { Meteor } from 'meteor/meteor';
 import { Restivus } from 'meteor/nimble:restivus';
 
 const ApiV1 = new Restivus({
-  apiPath: 'rest-api',
+  apiPath: 'api',
   version: 'v1',
   defaultHeaders: {
     'Content-Type': 'application/json',
--- a/core/server/api.js
+++ b/core/server/api.js
@@ -19,7 +19,7 @@ ApiV1.swagger = {
     swagger: '2.0',
     info: {
       version: '1.0.0',
-      title: 'Apinf API',
+      title: 'Admin API',
     },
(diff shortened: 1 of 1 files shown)


core/server/api.js
```diff
@@ -19,15 +19,7 @@ ApiV1.swagger = {
...
-      title: 'Apika API',
-      description: 'Apika REST API',
-      termsOfService: 'https://apika.digipalvelutehdas.fi/terms/',
-      contact: {
-        name: 'Apika team',
-      },
-      license: {
-        name: 'MIT',
-      },
+      title: 'Admin API',
```
jykae: @brylie make sure swagger.json is still valid if you remove most of the meta. Note also that currently REST API is not admin only. GET endpoints can be used by any logged in user.

brylie: Is the swagger.json *not* valid here?

http://swagger.io/specification/#infoObject

jykae: @brylie ok, tried also with validator, http://bigstickcarpet.com/swagger-parser/www/index.html Looks like it is valid.

Though if we call it "Admin API", we should restrict all endpoints 'admin' role only.
Is that requested?

brylie: That isn't the suggestion. I was just choosing a generic name that didn't use 'Apinf' or 'Apika'.

PR #207, opened 2017-01-06
Api LifeCycle Status Tag displays in Catalogue (table/grid)

### LINKED ISSUE #76: Show lifecycle status for API
# User story
> As an API user
I want to be able to see the lifecycle status of an API
so that I can determine how safe it is to start using an API

# Definition of done
* [ ] Once owner has added a lifecycle status for API, it needs to be shown in
  * [ ] Catalog card view
  * [ ] Catalog table view
  * [x] API profile

# Wireframe
1. API cards showing 5 lifecycle statuses:
cardapidesign
cardapidevelopment
cardapitesting
cardapiproduction
cardapideprecated

from @bajiat 's feedback, we should replace the tag for "private" APIs. The reason is that an API may or may not be private, but it would always have a lifecycle status. so that tag should be placed first in API card

2. API lifecycle status can be seen as colored labels in both card and table view.
From both view, the APIs can be filtered by one/more lifecycle statuses.
[future implementation suggestion: filter by private and public APIs]

apicatalognew

3. In API profile, the status appears on top of the page.
On transiting from one status to another, the time of change and transit from status appears as information (alt. can also be shown as tooltip on hover)

lifecycleapiprofile3

snittoor:
Closes #76

### DIFF
--- a/api_catalog/client/table/table.html
+++ b/api_catalog/client/table/table.html
@@ -17,6 +17,9 @@
                 {{ api.name }}
               </a>
             </h4>
+            <!-- Api LifeCycleStatus Label -->
+            {{> apiLifecycleStatus api=api }}
+            
             {{# unless api.isPublic }}
               <span class="label label-default label-align">private</span>
             {{/ unless }}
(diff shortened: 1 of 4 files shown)


bajiat:
Does this PR close issue #76?

Issue #205, opened 2017-01-05
Audit translation tokens


bajiat:
Go through the en. i18n.json file. Make a project-wide search and check that each token is used. If a token is not used/found, remove the token.

For reference, here is how we can get all i18n tokens in use in our software:

```
grep -iRoh '{{_ .*}}' * > all.html.i18n.txt
grep -iRoh 'TAPi18n\.__(.*)' * > all.js.i18n.txt
```
Last time some i18n tags with spaces before the underscore were not found with the search. You can modify the search if you like. Or then double check the missing ones manually.

manzapanza:
@bajiat I'm improving the regex to consider also these extra spaces that could be there. I checked few tokens manually, but now I'm working on a shell script to do everything, what do you think about?

PR #208, opened 2017-01-09
Audit translation tokens with a bash script

manzapanza:
Closes: #205

Run the script from root folder:
$ ./check-i18n-tokens.sh

For now, the script shows a list of i18n tokens that should be added on en.i18n.json file.

Next step:
- Show a list of i18n tokens that should be removed from en.i18n.json file

### DIFF
--- a/check-i18n-tokens.sh
+++ b/check-i18n-tokens.sh
@@ -3,21 +3,50 @@
 fileEnI18n="core/lib/i18n/en.i18n.json"
 
 echo "Tokens that should be added on en.i18n.json file:"
+echo ""
 
 # FIND TOKENS FROM HTML FILES
-grep -iRoh '{{ *_ .*}}' * --exclude="*.md" | grep -o "['\"].*['\"]" | sed -e "s@['\"]@@g" | while read -r token ; do
+tokensHtmlFiles=$(grep -iRoh '{{ *_ .*}}' * --exclude="*.md" | grep -o "['\"].*['\"]" | sed -e "s@['\"]@@g")
+for token in $tokensHtmlFiles; do
   if ! grep -q "$token" "$fileEnI18n"; then
     echo "$token"
   fi
 done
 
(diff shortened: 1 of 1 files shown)


brylie:
Very cool.

manzapanza:
Now the script shows a list of i18n tokens that should be removed from en.i18n.json file.

Next step:

- Improve the script that generate the two lists considering also schema names fields

check-i18n-tokens.sh
```diff
@@ -0,0 +1,52 @@
...
+
+fileEnI18n="core/lib/i18n/en.i18n.json"
+
+echo "Tokens that should be added on en.i18n.json file:"
+echo ""
+
+# FIND TOKENS FROM HTML FILES
+tokensHtmlFiles=$(grep -iRoh '{{ *_ .*}}' * --exclude="*.md" | grep -o "['\"].*['\"]" | sed -e "s@['\"]@@g")
+for token in $tokensHtmlFiles; do
+  if ! grep -q "$token" "$fileEnI18n"; then
```
brylie: Add comment describing these lines. We try to explain our code in plain English, just as we would to a novice programmer.

brylie: I.e. simply describe what these two lines are trying to achieve.

check-i18n-tokens.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "$token"
+  fi
+done
+
+# FIND TOKENS FROM JS FILES
+tokensJsFiles=$(grep -iRoh 'TAPi18n\.__(.*)' * --exclude="*.md" | grep -o "['\"].*['\"]" | sed -e "s@['\"]@@g")
+for token in $tokensJsFiles; do
+  if ! grep -q "$token" "$fileEnI18n"; then
```
brylie: Add simple, descriptive comment here.

PR #220, opened 2017-01-11
Bugfix/proxybackend form close validation

jykae:
Closes https://github.com/apinf/platform/issues/1935

### DIFF
--- a/proxy_backends/server/methods.js
+++ b/proxy_backends/server/methods.js
@@ -2,7 +2,6 @@ import { Meteor } from 'meteor/meteor';
 
 import { ProxyBackends } from '/proxy_backends/collection';
 
-
 Meteor.methods({
   deleteProxyBackend (proxyBackend, deleteFromMongoDB = true) {
     // Get umbrellaBackendId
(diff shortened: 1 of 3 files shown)


marla-singer:
@jykae Reviewing

marla-singer:
@jykae It shows sAlert if frontend prefix isn't unique. But it shows as well if i change Proxy after that

**Steps**
1. Select Proxy №1 and fill no unique value in **/proxy_frontend_prefix/** field
2. Watch on sAlert
3. Select another Proxy, don't change value in **/proxy_frontend_prefix/** field (and this value is unique for another Proxy)

**Expected**
Result will be saved
**Found**
Message about "no unique path"

If my steps are not clear, let me know

marla-singer:
@jykae Another question: does it need that fronted prefix must be unique for total proxies? Not in scope of one proxy?

marla-singer:
Well, looks like uniqueness must be totally.

proxy_backends/server/methods.js
```diff
@@ -34,4 +34,12 @@ Meteor.methods({
       }
     );
   },
+  uniqueFrontendPrefix (proxyBackend) {
+    const p = ProxyBackends.findOne({ 'apiUmbrella.url_matches.frontend_prefix': proxyBackend.apiUmbrella.url_matches[0].frontend_prefix });
```
marla-singer: Please rename this variable to sensible value

jykae:
```
@jykae Another question: does it need that fronted prefix must be unique for total proxies? Not in scope of one proxy?
```
Good note. We might have logical error on this one, as we previously had just one proxy. Checking only proxyBackends for current proxy would be enough.

marla-singer:
@jykae Let's pass this checking on a new issue?

jykae:
@marla-singer well, if you think so. I think I could check that also as part of this PR using "proxyId" on the selector as well, but it will go for tomorrow.

bajiat:
@marla-singer @brylie Can we merge / finalize this PR or should I move this to the next sprint?

marla-singer:
@bajiat I'm going to finalize it today

bajiat:
Cool!

marla-singer:
@bajiat Fix the little remarks, merge this branch and close issue

PR #222, opened 2017-01-11
Bugfix/bookmarks in catalog

### LINKED ISSUE #221: User cannot filter by bookmarks
1. Visit https://staging.apinf.io  
2. Sign in 
3. Go to Catalog  (default is 'Show all' )
4. Select and bookmark any API(s)
7.  Click on My Bookmarks  

# Expected Result
Displays all bookmarked APIs   

# Actual Result
Shows an empty page   

# Screenshot   

to be added

marla-singer:
Closes #221

### DIFF
--- a/api_catalog/client/api_catalog.js
+++ b/api_catalog/client/api_catalog.js
@@ -45,6 +45,8 @@ Template.apiCatalog.onCreated(function () {
   instance.subscribe('allApiLogo');
   // Subscribe to all users, returns only usernames
   instance.subscribe('allUsers');
+  // Subscribe to bookmarks of current user
+  instance.subscribe('userApiBookmarks');
 
   // Watch for changes in the sort and filter settings
   instance.autorun(() => {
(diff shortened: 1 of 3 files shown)


api_catalog/client/api_catalog.js
```diff
@@ -45,6 +45,8 @@ Template.apiCatalog.onCreated(function () {
   instance.subscribe('allApiLogo');
   // Subscribe to all users, returns only usernames
   instance.subscribe('allUsers');
+  // Subscribe to bookmarks of current user
+  instance.subscribe('myApiBookmarks', userId);
```
brylie: While we are here, can we rename this to `userApiBookmarks`? It would be consistent with other recent changes to the [userManagedApis](https://github.com/Digipalvelutehdas/APIKA/blob/develop/apis/collection/server/publications.js#L4) and [userManagedOrganizations](https://github.com/Digipalvelutehdas/APIKA/blob/develop/organizations/collection/server/publications.js#L20) publications.

marla-singer: Done

PR #224, opened 2017-01-11
meteor update, all possible packages without blaze

### LINKED ISSUE #212: Update project dependencies
Around every 1-2 sprints, we have been updating all project dependencies. Run `meteor update --all-packages` to get our dependencies updated to the latest compatible versions.

jykae:
Closes #212 

Updates meteor + all possible packages without "blaze" that caused error.
Issue is fixed in blaze but not yet released. Decided to postpone update of blaze package.


jykae:
@marla-singer @brylie please review

marla-singer:
@brylie @jykae I detected problem in behaviour

marla-singer:
@jykae @brylie 
I have two organization^

1. **Name**: New company **Site**: http://new-company.org **Slug**: new-company
2. **Name**: Production **Site**: http://production.com **Slug**: production

Click on **Production** organization and see profile "New company" although slug is"production". Refresh pageand see profile "Production"

jykae:
@marla-singer I am pretty sure it's not because of this PR.

marla-singer:
@jykae I repeated on develop branch and everything is okay.

jykae:
@marla-singer I propose you add separate issue about slugs updating, if you think we have something like that. I can't possibly see why updating packages could cause issue with slugs O_o

marla-singer:
@jykae No, no, slug is not a problem! 
I have two organizations in catalog. See below
joxi_screenshot_1484204981203
When I'm clicking on Yandex, I see as below



Refresh page, see that

When I'm clicking on Google so that

brylie:
@marla-singer have you tried a `meteor reset`? That way we could determine whether existing data might be a factor in this bug.

marla-singer:
@brylie `meteor reset` isn't a solution. This merge will appear in production site where we can't reset data. What will we do if existing data is a factor?

brylie:
> What will we do if existing data is a factor?

Lets first see if it is a factor, then we can determine what to do.

marla-singer:
@brylie Existing data isn't a factor

brylie:
I think this issue is related to subscription 'pollution'. In other words, there are several active subscriptions when viewing the Organization Profile, including a subscription to get all `userManagedOrganizations`. This means, the data returned to the Organization Profile may be that for a different organization than expected.

marla-singer:
@brylie Then why does this problem happen only on current branch?

brylie:
screenshot_20170112_124327

> Then why does this problem happen only on current branch?

Dunno. Do you have a different idea?

marla-singer:
>  Do you have a different idea?

Nope. If you think that now it doesn't matter, merge it

brylie:
Well, it looks like we can fix this issue soon. There might be a problem with our publications/subscriptions. I notice there is an `Organizations` subscription active on every route:

# Organization Profile
screenshot_20170112_125327

# Front page
screenshot_20170112_125536

# Organizations publication
It looks like the Organizations publication is defined globally in [organizations/collection/server/publications.js:18](https://github.com/Digipalvelutehdas/APIKA/blob/develop/organizations/collection/server/publications.js#L17-L18): 

screenshot_20170112_125840

brylie:
I pushed some refactored changes in that file, but am not sure how to handle the pagination pub/sub properly.

brylie:
@marla-singer can you please take a look at the Pagination pub/sub, to see if there is a way to limit it to only where it is needed?

.meteor/versions
```diff
@@ -73,15 +73,16 @@ juliancwirko:s-alert-stackslide@3.1.3
 kadira:blaze-layout@2.3.0
 kadira:dochead@1.5.0
 kadira:flow-router@2.12.1
-kurounin:pagination@1.0.15
+kurounin:pagination@1.0.19
```
marla-singer: @brylie Don't update this package

brylie: OK, we can open an enhancement to fix the Pagination publication/subscription.

marla-singer: I'm not sure how we can fix\imporve publication for Pagination. It's package feature
https://atmospherejs.com/kurounin/pagination#usage

brylie: This bug seems to have been present in `kuronin:pagination` since version 1.0.16. I opened an upstream bug report https://github.com/Kurounin/Pagination/issues/35

brylie: I pinned the package at 1.0.15 for now, as requested.

marla-singer: Good

brylie: Perhaps we can also pin the Blaze package, or should we try to upgrade Blaze again soon?

marla-singer: Let's try to upgrade Blaze when new version will be ready

brylie:
OK, I think this PR is ready now. I added a few changes to cleanup publications and make explicit queries in the Organization Profile template. Please review.

PR #233, opened 2017-01-13
Hotfix/users security

### LINKED ISSUE #231: Users list visible to anonymous users
The users management page can be accessed by anonymous users via URL. This displays all emails and usernames registered on the system.

# Task
Add sufficient security permissions to user related views and publications.

brylie:
Closes #231 

# Changes
- rename `allUsersUsernamesOnly` publication for explicitness
- make sure only admin users can access /users route
- bump version to 0.38.1

### DIFF
--- a/about/client/about.html
+++ b/about/client/about.html
@@ -28,7 +28,7 @@
               Apinf
             </dt>
             <dd>
-              0.38.0
+              0.38.1
             </dd>
             <dt>
               API Umbrella
--- a/api_catalog/client/api_catalog.js
+++ b/api_catalog/client/api_catalog.js
@@ -2,14 +2,14 @@ import { Meteor } from 'meteor/meteor';
 import { Template } from 'meteor/templating';
 import { Roles } from 'meteor/alanning:roles';
 import { FlowRouter } from 'meteor/kadira:flow-router';
-
 import { Apis } from '/apis/collection';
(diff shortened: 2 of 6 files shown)


about/client/about.html
```diff
@@ -28,7 +28,7 @@ <h2 class="modal-title" id="apiBacklogFormModalLabel">
               Apinf
             </dt>
             <dd>
-              0.38.0
+              0.38.1
```
marla-singer: Why do you change version?

brylie: For the time being, we use [Semantic Versioning](http://semver.org/). This is an incremental change, so I bumped the 'patch' version. Esentially, any time we add/change/remove code, we have a new version.

brylie: Since this hotfix is going into the `master` branch, we need to bump the version.

api_catalog/client/api_catalog.js
```diff
@@ -9,6 +9,7 @@ import { ApiBookmarks } from '/bookmarks/collection';
 Template.apiCatalog.onCreated(function () {
   // Get reference to template instance
   const instance = this;
+
```
frenchbread: Remove spaces here for eslint.

brylie:
@frenchbread fixed issues.

frenchbread:
@brylie 🕶  awesome

PR #239, opened 2017-01-16
Connect api from organization profile

### LINKED ISSUE #159: Connect API to an organization from organization profile
An organization manager needs to be able to connect APIs to the manager's organization. The manager can only do this if he/she also an API manager.

The organization profile needs to contain an option to Connect APIs to an organization and that option is only visible if the user is both manager of current organization and an API owner.

# Definition of done
* [ ] An organization manager can see a Connect APIs to an organization option on the organization profile, if the current user also has the role of API owner
* [ ] The available choices for APIs to connect to are API(s) the current user manages
* [ ] If the user selects an API and saves the changes, a connection between API and organization is made

# Wireframes

1. from Settings tab, organization manager can select from a dropdown APIs that he manage to connect to his organization

![altorganizationprofilesettings]
altorganizationprofilesettings2

(alt. design suggestion: a small button in the API view in Organization profile to connect APIs to the organization)
altorganizationprofile3

2. On selecting and saving an API, an S alert appears confirming the Organization Manager that his/her API is now connected to the organization and can be viewed from the Organization profile
altorganizationprofilesettingssalertadmin

marla-singer:
Closes #159


organization_apis/collection/server/publications.js
```diff
@@ -35,15 +35,42 @@ Meteor.publish('organizationApis', (slug) => {
...
     // Create an array of API IDs
     const apiIds = _.map(organizationApis, organizationApiLink => {
       // Get API ID from link document
       return organizationApiLink.apiId;
     });
 
+    // Get collection name for APIs collection
+    const collectionName = Apis.rawCollection().collectionName;
+    // Follow on changing of organizationAPis DB cursor
+    const handler = cursor.observe({
```
brylie: Why do we need lines 49-72?

organizations/client/profile/connect_api/connect_api.html
```diff
@@ -0,0 +1,50 @@
+<template name="connectApiToOrganizationModal">
```
brylie: Lets use the same form template that was created to connect an API to an Organization. That form was designed to be re-used for this purpose. Using the same form would reduce the lines of code we need to maintain.

marla-singer: I couldn't re-use form.  Can't remember quickly reasons but I had some problems with it

organizations/client/profile/profile.html
```diff
@@ -1,18 +1,29 @@
...
+                {{_ 'organizationProfile_smallText_apisCount' managedApisCount}}
+              </small>
+            </h3>
+          {{/ if }}
         </div>
-      {{/ if }}
-    {{/if}}
+        {{# if organization.currentUserCanEdit }}
+          <div class="col-md-9">
+            <button type="button" class="btn btn-default pull-right" id="connect-api"
```
brylie: The column markup may be unnecessary here (dividing the row into 3 and 9 columns). All that is probably needed is to place the button within the `h3` tag, as the first element, with the `pull-right` class.

organizations/client/profile/profile.js
```diff
@@ -18,14 +19,13 @@ Template.organizationProfile.onCreated(function () {
...
-
-      // Subscribe to Organization APIs documents
-      instance.subscribe('organizationApis', organizationSlug);
-
-      // Subscribe to Organization APIs documents
-      instance.subscribe('organizationApiLinksByOrganizationSlug', organizationSlug);
+      instance.subscribe('singleOrganization', organizationSlug);
     }
+    // Subscribe to Organization APIs documents
+    instance.subscribe('organizationApis', organizationSlug);
```
brylie: These should also be within the `if (organizationSlug) { ... }` block, since they rely on the organization slug to have a value.

organizations/client/profile/profile.js
```diff
@@ -76,3 +76,12 @@ Template.organizationProfile.helpers({
     return managedApisCount;
   },
 });
+
+Template.organizationProfile.events({
+  'click #connect-api': () => {
+    const organizationId = Organizations.findOne()._id;
```
brylie: This `findOne` should be made explicit, by passing in the `slug` value. We have had trouble with generic `findOne()` queries. When multiple subscriptions are active, it is possible for other organizations to be in the client data context, and `findOne()` might return an unexpected organization.

brylie: Also, be aware that `findOne()` can return `undefined`. In the case of an `undefined` result, we will get an error about accessing an attribute of undefined, when we try to access `_id`.

organizations/server/methods.js
```diff
@@ -0,0 +1,45 @@
...
+import { Meteor } from 'meteor/meteor';
+import { check } from 'meteor/check';
+import { Roles } from 'meteor/alanning:roles';
+
+// APINF collections import
+import { Apis } from '/apis/collection';
+import { OrganizationApis } from '/organization_apis/collection';
+
+Meteor.methods({
+  getNoBookedApis (organizationId) {
```
brylie: Change this method name to be more 'grammatical'. What is it trying to do?

organizations/server/methods.js
```diff
@@ -0,0 +1,45 @@
...
+    // Check arguments
+    check(organizationId, String);
+
+    const userId = this.userId;
+
+    // Find result for user with role
+    if (userId) {
+      // Find & group all booked apis
+      const OrganizationApisLink = OrganizationApis.find().fetch();
```
brylie: Use camel case, with first letter being lowercase.

organizations/server/methods.js
marla-singer: Return APIs that isn't connected to organization.

organizations/server/methods.js
brylie: This variable name should also be plural, since it references an array.

organizations/server/methods.js
```diff
@@ -0,0 +1,45 @@
...
+
+      const bookedApis = _.map(OrganizationApisLink, (module) => {
```
brylie: What is meant by 'booked'?

brylie: Also, what is meant by 'module'?

marla-singer: Booked == connected to organization

organizations/server/methods.js
```diff
@@ -0,0 +1,45 @@
...
+      if (!userIsAdmin) {
+        // limit selection by belong to current user
+        queryParams.managerIds = userId;
+      }
+
+      // Return cursor on no booked apis
+      return Apis.find(queryParams).fetch();
+    }
+
+    // Return empty result
```
brylie: Describe why we return an empty result here.

organizations/client/profile/profile.html
marla-singer: Bad solution. We use Bootstrap and I don't see any reasons to use frame
```html
<h3> APIs <button>..</button></h3>
```

organization_apis/collection/server/publications.js
marla-singer: Because of our super organizationApis module. When user connects API to organization, these is to insert a new record in orgnizationApis collection. APIs collection doesn't know about organizationApis collection and isn't updated reactive. MiniMongo has old data. APIs view block doesn't updated

organizations/client/profile/connect_api/connect_api.js
```diff
@@ -0,0 +1,50 @@
...
+
+Template.connectApiToOrganizationModal.helpers({
+  organizationApisCollection () {
+    // Return collection for autoform
+    return OrganizationApis;
+  },
+  noBookedApis () {
+    const instance = Template.instance();
+    // Return list of apis without organization
+    return instance.noBookedApis;
```
brylie: Instance data is not reactive. This may not behave as expected, since the method callback is asynchronous. `ReactiveVar` will give us a reactive data source for this reactive context (template helper).

organizations/server/methods.js
brylie: Ok, cool.

There is already an autoform options helper for this field. Lets consider modifying that code, rather than duplicating it as a method.

organizations/client/profile/connect_api/connect_api.js
marla-singer: I use `dataIsReady` variable

organizations/server/methods.js
brylie: OK, lets at least come up with a more natural sounding name. E.g. `linkedApis` and `getUnlinkedApis`.

organizations/client/profile/profile.html
brylie: Why is it bad? What do you mean by 'frame'?

organization_apis/collection/server/publications.js
brylie: What template/view specifically? I will try to reproduce locally.

organizations/client/profile/connect_api/connect_api.html
brylie: OK, lets figure out what those problems are, because the form was designed for re-use.

organizations/client/profile/profile.html
brylie: E.g. we use the same pattern in our [Proxy settings page](https://github.com/Digipalvelutehdas/APIKA/blob/develop/proxies/client/proxies.html#L6-L11)

- h1
  - button.pull-right
  - (text)

That markup is clean, and doesn't require any column classes for positioning. 

Is there some other problem with the solution?

organizations/client/profile/connect_api/connect_api.js
brylie: Right, it might just be redundant to imperatively control the `dataIsReady` variable when the reactive variable will automatically trigger the template helper to re-run.

All we need to do in the template is to check `{{# if noBookedApis }}`, and return undefined from the template helper when the reactive variable is not ready.

organizations/client/profile/profile.html
marla-singer: Using word `frame` I meant construction.
Bad solution is because later we will have filter icon and other icons. And you suggest to place all items in `h3` tag and specify class `pull-right`. `h3` tag will influence on filter icon size

organizations/client/profile/connect_api/connect_api.js
brylie: Basically, we can do the same thing with fewer lines of code, by properly embracing reactivity.

organization_apis/collection/server/publications.js
brylie: > APIs collection doesn't know about organizationApis collection and isn't updated reactive.

I don't understand why the `Apis` collection needs to 'update reactively'. The reactivity should come from proper publications and subscriptions relating to Organization APIs, and making sure the Organization APIs are fetched in a reactive context.

organizations/server/methods.js
marla-singer: It isn't a duplicate. Autoform options in your link returns **all** value.

organizations/client/profile/connect_api/connect_api.js
brylie: I opened PR #244 against this branch to demonstrate the idea.

brylie: If it won't work, this is OK. I just want to see if there is a slightly simpler approach.

organization_apis/collection/server/publications.js
marla-singer: Okay. Follow on steps:
1. Change my publication to old
2. Go to organization profile and click on button "Connect API"
3. Select API and click on Add

Found
Organization doesn't have added API but count is correct
joxi_screenshot_1484648971559

marla-singer: We need APIs collection reactive context. OrganizationAPis module doesn't have values to display in this part
joxi_screenshot_1484648971559 1

organizations/server/methods.js
brylie: Right, so lets see if we can modify that code, rather than creating a new method.

marla-singer: Organization is a new feature as the OrganizationApis.Now we don't have alternative

organization_apis/collection/server/publications.js
brylie: The subscriptions should get the right data reactively, because they use collection cursors:

screenshot_20170117_131900

As an aside, we don't need all three subscriptions, since two of them seem to do the same thing (looking at the comments).

brylie: Nevermind, just the comment needs to be changed.

brylie: In develop branch, when connecting APIs to an Organization via the API Profile, the Organizaiton Profile page workes correctly:

screenshot_20170117_133426

brylie:
At the least, lets just rename the method and variable names to be more sematic. We can open separate issues to see if this can be done more simply later.

organizations/client/profile/connect_api/connect_api.js
```diff
@@ -0,0 +1,52 @@
...
+import { Template } from 'meteor/templating';
+import { ReactiveVar } from 'meteor/reactive-var';
+
+// APINF collections import
+import { OrganizationApis } from '/organization_apis/collection';
+
+Template.connectApiToOrganizationModal.onCreated(function () {
+  const instance = this;
+  // Set status data is not ready
+  instance.dataIsReady = new ReactiveVar(false);
```
brylie: Remember to clean this up.

organizations/client/profile/connect_api/connect_api.js
```diff
@@ -0,0 +1,52 @@
...
+  instance.unlinkedApis = new ReactiveVar();
+  // Get organization id
+  const organizationId = Template.currentData().organizationId;
+
+  // Get data about apis without organization
+  Meteor.call('getUnlinkedApis', organizationId, (error, apis) => {
+    // Set status data is ready
+    instance.dataIsReady.set(true);
```
brylie: Remember to clean this up, if it is no longer used.

organizations/client/profile/connect_api/connect_api.js
```diff
@@ -0,0 +1,52 @@
...
+    const instance = Template.instance();
+    // Return list of apis without organization
+    return instance.unlinkedApis.get();
+  },
+  userCanConnectApi () {
+    const instance = Template.instance();
+    // Get count of free apis
+    return instance.unlinkedApis.get().length > 0;
+  },
+  dataIsReady () {
```
brylie: Remember to clean this up, if it is no longer used.

organizations/client/profile/profile.html
```diff
@@ -1,18 +1,25 @@
 <template name="organizationProfile">
-    {{# if Template.subscriptionsReady }}
-      {{> organizationProfileHeader organization=organization }}
-
+  {{# if Template.subscriptionsReady }}
+    {{> organizationProfileHeader organization=organization }}
+    <div class="container">
+      <div class="pull-right">
+        {{# if organization.currentUserCanEdit }}
```
brylie: Simply moving the button inside the `h3`, keeping `class="... pull-right..."` on the button, will improve this template:
- less code
- better alignment with the heading text

# Outside of `h3`
screenshot_20170117_151240

# Inside `h3`
screenshot_20170117_151321

```html
<template name="organizationProfile">
  {{# if Template.subscriptionsReady }}
    {{> organizationProfileHeader organization=organization }}
    <div class="container">
      {{# if managedApis }}
        <h3>
          {{# if organization.currentUserCanEdit }}
            <button type="button" 
              class="btn btn-default pull-right" 
              id="connect-api"
              data-toogle="tooltip" >
              <i class="fa fa-plus" aria-hidden="true"></i>
              {{_ 'organizationProfile_buttonText_connectApi' }}
            </button>
          {{/ if }}
          {{_ 'organizationProfile_bigText_apisCount'  }}
          <small>
            {{_ 'organizationProfile_smallText_apisCount' managedApisCount}}
          </small>
        </h3>
      {{/ if }}
      {{> apiCatalogGrid apis=managedApis }}
    </div>
  {{/if}}
</template>
```

brylie: Putting the button inside of the header element is consistent with other templates in our project. For example, the `organizationProfileHeader` template contains an `h1` with a nested `button`. These templates render on the same 'view'.

organizations/client/profile/profile.js
```diff
@@ -76,3 +78,21 @@ Template.organizationProfile.helpers({
...
+    // Get Organization, based on slug
+    const organization = Organizations.findOne({ slug });
+
+    // Check organization exist
+    if (organization) {
+      // Show modal with list of suggested apis and id of current organization
+      Modal.show('connectApiToOrganizationModal', { organizationId: organization._id });
+    } else {
+      // Otherwise throw error
+      throw new Meteor.Error('Organization document is undefined');
```
brylie: We will need to internationalize this error message, since it appears in the client.

organizations/client/profile/profile.html
marla-singer: @brylie Add filter icon before button

organizations/server/methods.js
```diff
@@ -0,0 +1,45 @@
...
+import { Meteor } from 'meteor/meteor';
+import { check } from 'meteor/check';
+import { Roles } from 'meteor/alanning:roles';
+
+// APINF collections import
+import { Apis } from '/apis/collection';
+import { OrganizationApis } from '/organization_apis/collection';
+
+Meteor.methods({
+  getUnlinkedApis (organizationId) {
```
brylie: We can probably go ahead and merge this solution for now. We may enhance it later.

marla-singer:
@brylie PR is not ready for review, i will ping you

organizations/client/profile/profile.html
marla-singer: Filter icon looks very big! 
joxi_screenshot_1484659689383

marla-singer: > Putting the button inside of the header element is consistent with other templates in our project.

Using wrong markup all-around doesn't turn wrong makeup to correct

organization_apis/collection/server/publications.js
brylie: I honestly believe there is a simpler solution here. For example, the *apis count* updates when connecting a new API to the organization:

peek 2017-01-17 15-40

Lets double check how the APIs Count gets its data, and see why the API Cards are not also getting the data reactively.

organizations/client/profile/profile.html
marla-singer: joxi_screenshot_1484660554905

brylie:
Cool, thanks.

organization_apis/collection/server/publications.js
marla-singer: @brylie APIs count looks at organizationApis. Of course it is updated! You insert a record to organizationApi collection on client-side DB. But Apis collection on client-side DB doesn't know that data was updated on organizationApis collection. And why Apis collection on client-side DB must be udated? No reason.

I really want to save code more simply. But it is impossible because APIs collection independence on organizationApis

brylie: Hm... what about returning an [array of cursors from the publication](http://docs.meteor.com/api/pubsub.html#Meteor-publish)?

organizations/client/profile/profile.html
marla-singer: @brylie I have issue about implement filter on Organization profile. So I will resolve our ~~holy war~~ question about markup another PR

organization_apis/collection/server/publications.js
marla-singer: Because publication run only on `onCreated` hook and isn't re-run

brylie: Another approach to consider: what would be the complexity if we were to subscribe to an array of APIs, based on the OrganizationApis documents in the client? E.g.

```
// In template autorun
// get OrganizationApis documents
// get array of API IDs
// subscribe to APIs collection passing in array of API IDs
```

marla-singer: I've solved this problem during 3 worked days

marla-singer: No problem, you can try this

marla-singer:
@brylie Fix remarks with variable names, comments, i18n tags

organizations/client/profile/profile.html
brylie: > Using wrong markup all-around doesn't turn a bad solution into a good

Why is the markup *wrong*?

marla-singer: @brylie Because "The `<h1>` to `<h6>` tags are used to define HTML headings."
http://www.w3schools.com/tags/tag_hn.asp
Button is not heading

marla-singer: Anyway, as I wrote above, I'll resolved markup question in another PR

marla-singer:
@brylie Let's merge it to I can continue working on Filter task

brylie:
OK, one moment. I have an initial re-work of the Organization APIs subscription, in a separate branch.

brylie:
Please see PR #253.

marla-singer:
@brylie Merge your branch. Please test it again and if it's okay, let's merge it

organizations/client/profile/connect_api/connect_api.html
```diff
@@ -0,0 +1,50 @@
...
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <i class="fa fa-times" aria-hidden="true"></i>
+          </button>
+          <h1 class="modal-title">
+            {{_ 'connectApiToOrganizationModal_modalTitle' }}
+          </h1>
+        </div>
+        <div class="modal-body">
+          {{# if dataIsReady }}
```
brylie: Do we need the `dataIsReady` template helper any more?

marla-singer: Yes. When `unlinkedApis` is empty array, spinner switches forever

brylie: OK, can we return `undefined` from `unlinkedApis` rather than an empty array?

marla-singer: As for me, `dataIsReady` and `unlinkedApis` have different semantic
- If data is ready then these shows form else these shows spinner
- If unlinkedApis has values then these shows drop-down list with values else these shows text "Every apis are linked"

brylie:
Removed 'ready for review', waiting for feedback in PR #260.

brylie:
@marla-singer, I made three more commits. Basically, I renamed the `getCurrentUserUnlinkedApis` method, and return `undefined` when no unlinked APIs are available. Returning `undefined`, instead of an empty array, means we don't need the `dataIsReady` template helper (fewer moving parts).

brylie:
Please review, and merge if you approve.

marla-singer:
Well, maybe spinner is not needed in this case :eyes:

marla-singer:
Approve PR :+1: As PR authorI can't do it via GitHub feature
Let's merge it

brylie:
> Well, maybe spinner is not needed in this case

I understand your reasoning, and agree with your idea about the helpers being distinct. I am willing to add it back, if removing it causes bugs. Lets keep our eyes open for bugs here.

PR #244, opened 2017-01-17
Simplify template helpers using reactive var

brylie:
Examples of suggested changes to PR #239


marla-singer:
@brylie In case when `noBookedApis` has empty list as result and spinner switches on forever

Issue #236, opened 2017-01-15
Add eslint check to travis build


mauriciovieira:
Currently, there is no static code check in the [continuous integration](https://travis-ci.org/Digipalvelutehdas/APIKA). It is desirable to add a eslint check during pull requests because that will help avoiding the introduction of new bugs.

# Definition of done
* [x] Add eslint check to continuous integration (#237)
* [ ] Make eslint to check all directories
  * [x] about (#237)
  * [x] api_catalog (#237)
  * [x] api_keys (#237)
  * [x] apis
    * [x] apis/client (#282)
    * [x] apis/collection (#316)
    * [x] apis/flags (#317)
    * [x] apis/logo (#318)
    * [x] apis/server (#252)
  * [x] apis_authorization (#281)
  * [ ] backlog (#341)
  * [ ] bookmarks (#350)
  * [ ] branding (#351)
  * [ ] core (#352)
  * [ ] dashboard (#353)
  * [x] docker
  * [ ] documentation (#354)
  * [ ] elasticsearch (#355)
  * [ ] feedback (#356)
  * [ ] feedback_votes (#356)
  * [ ] home
  * [ ] metadata
  * [ ] monitoring
  * [x] node_modules
  * [x] organization_apis (#260)
  * [x] organizations (#260)
  * [ ] proxies
  * [ ] proxy_backends
  * [ ] public
  * [ ] ratings
  * [ ] scripts
  * [ ] search
  * [ ] settings
  * [ ] setup
  * [ ] status
  * [ ] users

mauriciovieira:
@bajiat I am working on this. I talked to @brylie Friday and we think it's a good idea.

mauriciovieira:
I am doing this 'locally', and it seems we'll have a long work to do. According to https://travis-ci.org/mauriciovieira/platform/builds/192108795, there are _1656 problems (1358 errors, 298 warnings)_  right after the latest Jan 12th commit (https://github.com/mauriciovieira/platform/commit/0c6b28f7d00ee1eb9328e0341fa829cdd095e827) 

What I propose is to not add all directories in a single step, but one directory at a time, and fix the lints progressively.

mauriciovieira:
And I am pretty sure this step will help us not to introduce new bugs to the code. We already have a https://docs.meteor.com/packages/audit-argument-checks.html to the eslintrc, and it found a lot of non-checked arguments in the code.

mauriciovieira:
[...]

PR #252, opened 2017-01-18
Add apis/server dir to lint

mauriciovieira:
This PR is part of #236 endeavor.


mauriciovieira:
Please review and merge if you approve.

brylie:
One problem I just realized about this approach, is that now my IDE won't lint the files in .eslintignore. I understand this is by design, but we need to make sure all developers are on the same page about how to lint the project going forward.

brylie:
@bajiat we need to let all developers know about this PR. Going forward, whenever we touch an existing JavaScript file in our project, we need to make sure it is removed from the .eslintignore file and passes lint. We are going to work through our project to 100% passing lint.

PR #253, opened 2017-01-18
Connect api profile subscription change

brylie:
Related to PR #239

This is a work in progress, but demonstrates an alternative approach for getting Organization API documents.

### DIFF
--- a/organizations/client/profile/profile.js
+++ b/organizations/client/profile/profile.js
@@ -4,6 +4,7 @@ import { FlowRouter } from 'meteor/kadira:flow-router';
 import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
 import { TAPi18n } from 'meteor/tap:i18n';
 import { sAlert } from 'meteor/juliancwirko:s-alert';
+import _ from 'lodash';
 
 // APINF collections import
 import { Organizations } from '/organizations/collection/';
@@ -31,15 +32,18 @@ Template.organizationProfile.onCreated(function () {
       const organization = Organizations.findOne({ slug: organizationSlug });
 
       // Get all Organization API links
-      const apiLinks = OrganizationApis.find({ organizationId: organization._id });
+      const apiLinks = OrganizationApis.find({ organizationId: organization._id }).fetch();
 
-      // Create an array of Organization API IDs
-      const apiIds = _.map(apiLinks, function (apiLink) {
-        return apiLink.apiId;
-      });
+      // Make sure there is at least one Organization API
+      if (apiLinks.length > 0) {
+        // Get all Organization API IDs
+        const apiIds = _.map(apiLinks, function (apiLink) {
+          return apiLink.apiId;
+        });
 
-      // Subscribe to Organization APIs documents
-      instance.subscribe('apisById', apiIds);
+        // Subscribe to Organization APIs documents
+        instance.subscribe('apisById', apiIds);
+      }
     }
   });
 });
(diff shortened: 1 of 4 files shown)


marla-singer:
Did you run it? I have error on browser console
`TypeError: Cannot read property '_id' of undefined` when I opened Organization profile

brylie:
It is a work in progress. That error is probably easy to resolve. It is intended to show a different approach.

brylie:
OK, I worked out the bugs. Please review.

marla-singer:
The page blinks when the connection is done. Well, maybe the code is simple but behaviour is not pretty good

organizations/client/profile/profile.js
```diff
@@ -23,11 +25,27 @@ Template.organizationProfile.onCreated(function () {
...
+      const organization = Organizations.findOne({ slug: organizationSlug });
+
+      if (organization) {
+        // Get all Organization API links
+        const apiLinks = OrganizationApis.find({ organizationId: organization._id }).fetch();
+
+        // Make sure there is at least one Organization API
+        if (apiLinks.length > 0) {
+          // Get all Organization API IDs
+          const apiIds = _.map(apiLinks, function (apiLink) {
```
marla-singer: Use organization.apis() helper to get managed apis

brylie: I can't get the APIs until we subscribe to them. In this context, I am building an array of API IDs in order to subscribe to the Organizaiton APIs.

brylie: In other words, `organization.apis()` will only work if the subscription is already active, and APIs are available in the client.

marla-singer: Then we have code duplicate 
https://github.com/Digipalvelutehdas/APIKA/blob/develop/organization_apis/collection/helpers/organizations.js#L12-L23

marla-singer: Save this value in the instance variable and then use it in managedApis helper instead of `organization.apis()`

marla-singer: And `managedApisCount` can be as `instance.managedApis.length`

brylie: Well, the collection helper is still useful. We can leave it.

marla-singer: No, don't delete it. Just don't use in this case

brylie: The collection helper just needs a matching subscription on the client side. E.g. the `organization.apis()` helper relies on the following collections
- `organizationApiLinksByOrganizationSlug`
- `apisById` (using API IDs from the OrganizationApis documents

brylie: I will simplify the profile.js

marla-singer: Hm, right. I've got

brylie: OK, I removed the unnecessary helper from profile.js, and used the `organization.apis` helper in the profile.html

organizations/client/profile/profile.js
```diff
@@ -23,11 +25,27 @@ Template.organizationProfile.onCreated(function () {
...
+            return apiLink.apiId;
+          });
+
+          // Subscribe to Organization APIs documents
+          instance.subscribe('apisById', apiIds);
```
marla-singer: Get semantic name for publication

brylie: What do you recommend?

marla-singer: My English skill is so bad. These are obvious to use `organizationApisByIds`

brylie: Cool. One thing to realize, is this takes *any* array of API IDs. In other words, it is not specific to the Organizations feature, and might be used in other contexts.

marla-singer: Okay, let's name as `apisByIds`

PR #269, opened 2017-01-20
Using helper for Edit button

### LINKED ISSUE #214: User without permission can see Edit button on organization profile
On the Organization Profile, the Edit button is visible to anonymous users:

screenshot_20170112_143537

### Steps
1. As an anonymous user, navigate to an organization
2. Click on Edit button (place is near organization name)
3. Try to edit data

### Found
- Form is opened
- After editing and saving nothing is happened 
- These is error on browser console about "Access denied [403]"

### Expected
- Don't show Edit button to any users, only for limited audience

# Task
Make sure the Edit button only displays for Organization Managers or Admin users. there may already be an `organization.currentUserCanEdit` helper, otherwise we may need to create it.

marla-singer:
Closes #214

### DIFF
--- a/organizations/client/profile/header/header.html
+++ b/organizations/client/profile/header/header.html
@@ -10,9 +10,11 @@
           <div class="row">
             <h1 id="organization-name">
               {{ organization.name }}
-              <button id="edit-organization" class="btn btn-default btn-xs">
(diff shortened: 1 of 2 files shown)


marla-singer:
@brylie Could you please reviewing this little PR?

Issue #263, opened 2017-01-19
User unable to Get API key in nightly


bajiat:
# Steps to reproduce
1. Log in to nightly.apinf.io as admin
2. Connect one of the available APIs to either one of the available API Umbrella proxies.
3. Press Get API key

# Expected result:
API key is successfully retrieved and can be copied from API profile

# Actual result:
There is a success message "API key successfully retrieved", but UI is showing Processing... and gets stuck for at least 5-10 minutes (after which I lost patience and refreshed the page. If you refresh the page, you can see GET API key button again.)

See screenshots from Comments

bajiat:
<img width="888" alt="get api key 1" src="https://cloud.githubusercontent.com/assets/11626388/22110691/4a4f486a-de65-11e6-9c33-c4e4eec8fdc1.png">

<img width="894" alt="get api key 2" src="https://cloud.githubusercontent.com/assets/11626388/22110702/50b01770-de65-11e6-9917-3b35fd1586d6.png">

<img width="889" alt="get api key 3" src="https://cloud.githubusercontent.com/assets/11626388/22110709/56605040-de65-11e6-95f9-f79aa8848b15.png">

bajiat:
@marla-singer Could you take this task as well?

marla-singer:
@bajiat Tested it on nightly.apinf.io and it works as expected.

bajiat:
@marla-singer So it was fixed by #272 ?

marla-singer:
@bajiat Yep.

saralavanip:
@bajiat , verified this issue on https://nightly.apinf.io v 0.40.0  and it is reproducible when API connected to 'Second nightly' proxy.

PR #272, opened 2017-01-23
Fix import of ApiKeys collection

marla-singer:
Some files had incorrect APiKeys collection import.
Maybe it will resolve issue #263


mauriciovieira:
I wonder if it wouldn't be something caught by eslint.

brylie:
Good find!

PR #274, opened 2017-01-23
Add handler for Clear all button

### LINKED ISSUE #273: "Clear all" button doesn't work as expected
I tested APIs filter on nightly.apinf.io. Found that "Clear all" button doesn't work as expected.
joxi_screenshot_1485158545531

marla-singer:
Closes #273

### DIFF
--- a/organizations/client/profile/profile.js
+++ b/organizations/client/profile/profile.js
@@ -119,7 +119,7 @@ Template.organizationProfile.events({
       sAlert.error(message);
     }
   },
-  'click [type=reset]': () => {
+  'click #reset-filter-options': () => {
     // Delete query parameter
     FlowRouter.setQueryParams({ lifecycle: null });
   },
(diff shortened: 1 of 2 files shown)


organizations/client/profile/profile.js
```diff
@@ -119,4 +119,8 @@ Template.organizationProfile.events({
       sAlert.error(message);
     }
   },
+  'click [type=reset]': () => {
```
brylie: Lets add an `id` to the reset button, so that we can refer to it by name:

```js
' click #reset-filters` ...
```

brylie:
Small change requested, for readability. Similar to [using explanatory variables](https://github.com/ryanmcdermott/clean-code-javascript#use-explanatory-variables), our HTML can use explanatory IDs, so that we understand what our code is referencing.

organizations/client/profile/profile.js
marla-singer: I thought it isn't the important element to have got id

marla-singer: @brylie Done

brylie:
Ready for review?

marla-singer:
@brylie Yep

PR #280, opened 2017-01-23
Remove record from organizationApis

### LINKED ISSUE #279: OrganizationApis module is not updated when user deletes API
OrganizationApis module is not updated when user deletes API

### Step

1. Create an Organization and an API, navigate to Organization
2. Pay attention that Organization profile doesn't have connected API and doesn't have the title "**APIs**" as expected. There are empty area in blue frame on screenshot (as expected)
joxi_screenshot_1485186270442 1
3. Connect a created API to created organization. 
4. The APIs title has appeared on Organization profile as expected
joxi_screenshot_1485186270442
5. Go to API page and delete it
6. Navigate to Organization

### Found
- Title **APIs (0 apis)** 
joxi_screenshot_1485186852320
### Expected
- Empty area as before connection
### Reasons
- It happens because OrganizationApis collection has record about deleted API. You can see it on screenshot

marla-singer:
Closes #279

### DIFF
--- a/apis/server/methods/delete.js
+++ b/apis/server/methods/delete.js
@@ -8,6 +8,7 @@ import { ApiMetadata } from '/metadata/collection';
 import { DocumentationFiles } from '/documentation/collection/collection';
 import { Feedback } from '/feedback/collection';
 import { MonitoringSettings, MonitoringData } from '/monitoring/collection';
+import OrganizationApis from '/organization_apis/collection';
 import { ProxyBackends } from '/proxy_backends/collection';
 
 Meteor.methods({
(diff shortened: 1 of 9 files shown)


apis/server/methods/delete.js
```diff
@@ -79,4 +83,17 @@ Meteor.methods({
     // Remove monitoring settings collection
     MonitoringSettings.remove({ apiId });
   },
+  removeOrganization (apiId) {
```
brylie: The name of this function implies that we are removing an organization ("remove organization"). However, the code looks like we are removing an **organization-api link**, the linking object between an organization and an API. Lets rename this function to match what it actually does, e.g. `removeOrganizationApiLink`.

Reference
- Clean Code Javascript: [Function names should say what they do](https://github.com/ryanmcdermott/clean-code-javascript#function-names-should-say-what-they-do)

brylie:
Made a small suggestion about the method name. Also, please merge conflicts.

marla-singer:
@brylie Done

PR #295, opened 2017-01-25
Fix links

### LINKED ISSUE #294: Fix links in About
Some links are incorrect on About modal.
Specifically: 
[https://apinf.org](https://apinf.org) should be [http://apinf.org](http://apinf.org)
[https://apinf.org/docs](https://apinf.org/docs) should be [http://apinf.org/docs](http://apinf.org/docs)
We can check if it possible to have these links with https, but for now we need to make sure users are getting to working website.

nnn:
Closes #294

### DIFF
--- a/about/client/about.html
+++ b/about/client/about.html
@@ -45,8 +45,8 @@
               {{_ "aboutApinf_homepage" }}
             </dt>
             <dd>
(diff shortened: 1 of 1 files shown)


nnn:
@Digipalvelutehdas/apika-tiimi please review.

PR #302, opened 2017-01-26
Organization managers inherit API manager rights for connected APIs

### LINKED ISSUE #133: Organization managers inherit API manager rights for connected APIs
When an API gets connected to an organization, each of the organization managers for that particular organization should inherit API manager rights for any APIs that are connected to the organization. 

# Definition of done
* [ ] An organization manager has the same rights as the original API manager to edit, delete the API after an API has been connected to an organization.

nnn:
Closes #133


organizations/server/methods.js
```diff
@@ -47,7 +48,7 @@ Meteor.methods({
     return undefined;
   },
   addOrganizationManagerByEmail (manager) {
-    // Make sure organizationId is a string
+    // Make sure manager (with organizationId and email) is an Object
```
brylie: > Make sure manager (with organizationId and email)

Add subsequent checks for the expected object properties:

```js
check(manager.organizationId, String);
check(manager.email, String);
```

brylie:
How are we handling the situation where:

1. `username1` is added to `organizaiton1` managers
2. `api2` is added to organization APIs

In other words, how can we *dynamically* check whether an user has organization manager status?

I think the answer might come from [Role groups](https://github.com/alanning/meteor-roles#roles-groups), where each organization has its own group and users are given roles within the group.

nnn:
Thanks @brylie for pair-programming session, it was quite a journey to make it work :)
@marla-singer or @frenchbread could you please review this PR as well.

frenchbread:
Awesome 🕶

PR #304, opened 2017-01-26
Private api in organization profile

### LINKED ISSUE #301: Anonymous users can view 'Private' APIs listed in 'Organization catalog'
# Steps  
1. Visit https://staging.apinf.io (version 0.39.0)   
2. Login as admin 
3. Add API 
4. Go to 'My APIs' and click on above API added 
5. Click on Settings 
6. Change the visibility of API from to 'Private'  
7. Add Organization 
8. Connect the API to above Organization  
9. Logout 
10. Again visit https://staging.apinf.io  
11. Navigate to 'Organizations catalog' and select the above organization  

# Expected Result   
-  Private APIs should not be available  in organization catalog for anonymous users. 
-  Only admin or the owner of that particular private labeled API can view in organization  catalog  

# Actual Result  
-  Private APIs  available  in organization catalog for anonymous users. 
-  Select this private API , displays error: 403 'You don't have permissions to view this page' 
- Click on 'Return to previous page' is unresponsive   

# Screenshot   
anonymous_users_view_private_apis_org_catalog1

anonymous_users_view_private_apis_org_catalog

marla-singer:
Closes #301 
- APIs tab is shown for admin and manager of current organization all time
- APIs tab isn't shown if organization doesn't have public APIs
- Content of APIs tab has an own template. 
- The new publications: for **all** apis and for **user available** apis


brylie:
While this solution works, it is quite complicated and duplicates significant code. Also, the helper functions do much more than their name implies.

Lets try a simpler solution, with more focused and semantic collection helper names.

# References
- Clean Code JavaScript - [Function names should say what they do](https://github.com/ryanmcdermott/clean-code-javascript#function-names-should-say-what-they-do)
- Clean Code JavaScript - [Functions should do one thing](https://github.com/ryanmcdermott/clean-code-javascript#functions-should-do-one-thing)
- Clean Code JavaScript - [Remove duplicate code](https://github.com/ryanmcdermott/clean-code-javascript#remove-duplicate-code)

organization_apis/collection/helpers/organizations.js
```diff
@@ -21,6 +61,13 @@ Organizations.helpers({
     // Return a count of Organiztion-API Links
     return OrganizationApis.find({ organizationId: this._id }).count();
   },
+  availablePublicApis () {
+    // Get array of managed apis which available for current user
+    const apis = this.apis();
+
+    // Return length of array
```
brylie: Either this comment is incorrect or the code is not doing what you expect. It seems like this returns a boolean: `true` if `apis.length` is greater than 0

organization_apis/collection/helpers/organizations.js
```diff
@@ -21,6 +61,13 @@ Organizations.helpers({
     // Return a count of Organiztion-API Links
     return OrganizationApis.find({ organizationId: this._id }).count();
   },
+  availablePublicApis () {
```
brylie: The name of this helper may be misleading. Is this returning the APIs or a **count** of APIs?

marla-singer: it returns true if the organization has public APIs to show for anonym or ordinary users

organizations/client/profile/profile.html
```diff
@@ -19,7 +19,7 @@
                 </button>
               {{/ if }}
             </div>
-            {{# if organization.apisCount }}
+            {{# if organization.availablePublicApis }}
```
brylie: Given that this helper is checking "if organization has APIs", lets change the collection helper name to `hasApis`.

brylie:
This solution should probably also involve one or more publications. The reason being, the client should not receive any data they are not entitled to access. In client-side code, collection helpers are relying on subscribed data. Rather than overpublishing data and filtering on the client, we should have a better-designed publication/subscription here.

organization_apis/collection/helpers/organizations.js
```diff
@@ -9,10 +11,48 @@ Organizations.helpers({
...
+      // Placeholder for filtering
+      let filtering;
+
+      // Set filters
+      // Case: Registered users
+      if (userId) {
+        // Case: user is admin or manager of organization
+
+        // Get all managed organizations fo current user
+        const userIsOrganizationManager = Organizations.find({ managerIds: userId }).count();
```
brylie: Consider the statement "if user is organization manager". It is basically a yes/no (true/false) statement. It looks like we are actually dealing with a count of managed organizations here, so the variable name should reflect its purpose.

- [Use meaningful variable names](https://github.com/ryanmcdermott/clean-code-javascript#use-meaningful-and-pronounceable-variable-names)
- [Use the same vocabulary for the same type of variable](https://github.com/ryanmcdermott/clean-code-javascript#use-the-same-vocabulary-for-the-same-type-of-variable)

organizations/client/profile/profile.html
marla-singer: Done

organization_apis/collection/helpers/organizations.js
marla-singer: Change comment

organization_apis/collection/helpers/organizations.js
marla-singer: Renamed helper

organization_apis/collection/helpers/organizations.js
marla-singer: There are tow ways:
1. Rename variable and it will be as
```javascript
// Get all managed organizations fo current user
const managedOrganizationsOfCurrentUser = Organizations.find({ managerIds: userId }).count();
// Check if user is admin
const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
if (userIsAdmin || managedOrganizationsOfCurrentUser > 0) {
...
}
```
2. Change logic
```javascript
// Get all managed organizations fo current user
const userIsOrganizationManager = Organizations.find({ managerIds: userId }).count() > 0;
// Check if user is admin
const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
if (userIsAdmin || userIsOrganizationManager) {
...
}
```

I prefer the second way because it doesn't matter how I get the value `userIsOrganizationManager`

marla-singer:
> This solution should probably also involve one or more publications. The reason being, the client should not receive any data they are not entitled to access. In client-side code, collection helpers are relying on subscribed data. Rather than overpublishing data and filtering on the client, we should have a better-designed publication/subscription here.

@brylie I will work on it

organization_apis/collection/helpers/organizations.js
brylie: I like the second approach too, because both `userIsAdmin` and `userIsOrganizationManager` are basically questions, boolean data type.

organizations/client/profile/header/header.js
```diff
@@ -1,12 +1,34 @@
...
 
+import OrganizationLogo from '/organizations/logo/collection/collection';
+
 Template.organizationProfileHeader.onRendered(function () {
   // Assign resumable browse to element
   OrganizationLogo.resumable.assignBrowse(this.$('#organization-file-browse'));
 });
 
+Template.organizationProfileHeader.helpers({
+  currentUserCanView () {
```
brylie: This should be done in the route, redirecting to a 403 not authorized page if user is not allowed to view. Check out how we handled private APIs.

brylie: Rename this helper to `currentUserCanEdit` and use it only to show and hide the *Connect API button**. That might mean we have to refactor the APIs list template, moving the button elsewhere.

marla-singer: Nope. this helper need to show or not APIs tab

organizations/client/profile/header/header.js
```diff
@@ -1,12 +1,34 @@
...
+    // Check user's role on admin
+    const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
+    const userIsOrganizationManager = organization.managerIds.indexOf(userId) > -1;
+
+    // Show APIs tab if user is admin or manager of current organization
+    if (userIsAdmin || userIsOrganizationManager) {
+      return true;
+    }
+    // Show APIs tab if public apis are available otherwise don't show
+    return organization.hasPublicApis();
```
brylie: Remove this line from the `currentUserCanManage` funciton. Instead, use it directly in the HTML template:

```html
{{# if organizaiton.hasPublicApis }}
  <!-- Show APIs list template -->
{{/ if }}
```

We are overworking this function, causing it to break the single purpose principle.

organization_apis/collection/helpers/organizations.js
```diff
@@ -9,10 +11,48 @@ Organizations.helpers({
   apis () {
```
brylie: For simplicity, this helper should just return **all** APIs related to an organization, regardless of user role or public/private.

Lets think at a higher level about the `organizations` collection helpers API. What would be a good structure?

brylie: For example:

- organization
  - apis - *all* APIs (regardless of public or private)
  - publicApis - only public APIs

brylie: Checking for user permissions is a separate function/helper:

- organization
  - currentUserCanManage - checks if user is Manager or Admin

marla-singer: We don't need a helper to returns **all** APIs related to an organization, regardless of user role or public/private. Because we have an user role and public/private apis. 

Now I created an own template for APIs tab. The next step will be integrate paginate package in this template as we have in APIs and Organizations catalog

marla-singer: Exactly! We **already** have helper `currentUserCanEdit`for organization. I will rename it to `currentUserCanManage`

brylie: We can use the `organization.currentUserCanManage` helper to control the subscription on the client side. It would toggle the subscription between:

```
if (organization.currentUserCanManage) {
  instance.subscribe(`allOrganizationApis`);
} else {
  instance.subscribe('organizationPublicApis');
}
```

Likewise, in the template helper:

```js
if (organization.currentUserCanManage) {
  return organization.apis();
} else {
  return organization.publicApis();
}
```

marla-singer: Don't forget about filter on this page :(

marla-singer: But I see your point

brylie:
Cool, lets refactor this code so it has cleaner separation, and a good high level API.

apis/collection/server/publications.js
```diff
@@ -9,22 +11,66 @@ Meteor.publish('userManagedApis', function () {
...
+  // Case: Registered users
+  if (userId) {
+    // Case: user is manager of APIs or without APIs
+
+    // Select available managed apis of current organization
+    filteredApis = {
+      _id: { $in: apiIds },
+      $or: [
+        { isPublic: true },
+        { managerIds: userId },
```
brylie: We still have [too many levels of abstraction](https://github.com/ryanmcdermott/clean-code-javascript#functions-should-only-be-one-level-of-abstraction) here. In other words, we are not only checking for `publicApisByIds` here, but we are also checking whether an User is in the `managerIds` array.

As the name connotes, this function should only return "public APIs (`isPublic: true`) that are in an array of API IDs".

organization_apis/collection/helpers/organizations.js
```diff
@@ -21,6 +22,13 @@ Organizations.helpers({
     // Return a count of Organiztion-API Links
     return OrganizationApis.find({ organizationId: this._id }).count();
   },
+  hasPublicApis () {
```
brylie: This function name may also be misleading. It reads like "the organization has public APIs", but instead returns a count of all APIs (`Apis.find().fetch()`). 

It is important that we are clear and careful that code does what we expect, and is easy to understand just by reading function and variable names.

organizations/client/profile/apis/apis.html
```diff
@@ -0,0 +1,37 @@
+<template name="organizationApis">
+  <!-- dropdown class set relative position -->
+  <div class="block-header">
+    <div class="pull-right">
+      <!-- Filter button -->
+      {{# if organization.hasPublicApis }}
```
brylie: This will return true if there are *any* API documents in the client subscription context. In other words, if there are any subscriptions active that return a cursor of one or more APIs, this will return true. It does not guarantee that the API belongs to the organization.

organizations/client/profile/header/header.js
brylie: Checking whether the "current user can view the organization" should be done at the router level.

brylie: I am nut sure if/why we need to show or hide the APIs tab. Also, why do we have an APIs tab? I thought we are displaying Organization APIs in a row/block on the Organization Profile.

organizations/client/profile/header/header.js
```diff
@@ -1,12 +1,26 @@
...
+Template.organizationProfileHeader.helpers({
+  currentUserCanView () {
+    const organization = Template.currentData().organization;
+
+    // Show APIs tab anyway if user is admin or manager of current organization
+    if (organization.currentUserCanManage()) {
+      return true;
+    }
+    // Show APIs tab if public apis are available otherwise don't show
+    return organization.hasPublicApis();
```
brylie: This helper is unnecessary and does more than the name connotes.

We can perform this check directly in the template, without needing a template helper. E.g.

```html
{{# if organization.hasPublicApis }}
  <!-- show the APIs row/block -->
{{/ if }}
```

organizations/client/profile/profile.html
```diff
@@ -3,47 +3,19 @@
     {{> organizationProfileHeader organization=organization }}
     <div class="container dropdown">
       <div class="tab-content">
+        <!-- APIs tab -->
```
brylie: Is there a wireframe, or design discussion about moving this to a tab? I just don't recall the decision.

organizations/collection/permissions.js
```diff
@@ -26,7 +26,7 @@ Organizations.allow({
     const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
 
     // Only admin/owner can update
-    if (userIsAdmin && organization.currentUserCanEdit()) {
+    if (userIsAdmin && organization.currentUserCanManage()) {
```
brylie: This should be a logical `or`. Right now, the code basically works like:

> If the user has Admin role AND they can manage the organization

Whereas, it should only check that the user can manage the organization:

> If the user has Admin role OR they are an Organization manager 

Further, we can check whether a user has an admin role within the `organization.currentUserCanManage` helper.

organizations/logo/client/resumable.js
```diff
@@ -57,7 +57,7 @@ Meteor.startup(() => {
 
     // Get organization
     const organization = Organizations.findOne();
-    if (organization && !organization.currentUserCanEdit()) {
+    if (organization && !organization.currentUserCanManage()) {
```
brylie: Refactor `organization.currentUserCanManage` to check whether user has Admin role OR is an Organization manager. I.e. BOTH Admin users and Organization managers can manage organizations.

brylie:
This is getting closer, but there are still substantial improvements that can be made. We need to focus on code readability and semantics, and not add too many layers of abstraction into functions where it isn't needed.

organizations/client/profile/profile.html
brylie: The [related bug report](https://github.com/Digipalvelutehdas/APIKA/issues/301) only mentions permissions related to public/private APIs, so the UI changes are probably not necessary here.

marla-singer: Wireframe for "Showing organization APIs on organization page "
How you can see we have two tabs: APIs and Settings
https://github.com/Digipalvelutehdas/APIKA/issues/74

organization_apis/collection/helpers/organizations.js
marla-singer: It returns all APIs from client-side collection

apis/collection/server/publications.js
marla-singer: Of course, APIs managers must see private API in organization profile. It's not "too many levels of abstaction here". It is just a query to database.
The name `availableApisForCurrentUser` connotes that this function should return all apis which current user can see with due regard his role and private/public state of APIs?

organizations/client/profile/header/header.js
marla-singer: If an organization doesn't have any APIs or has only private API then I hide APIs tab otherwise It looks confused for users (see example on link)
https://nightly.apinf.io/organizations/apinf

And I sure that we must use two different helpers - 
1. currentUserCanView needs to show or not content of APIs tab when organization doesn't have APIs or have only private API
      - if you are admin or manager, you must see tab because it has "Connect API to organization" button and you should manage private API
      - if you are an ordinary user, there are not to show APIs tab with an empty page
2. hasPublicApis needs to show or not label "APIs (apis 1)" and filter icon

organizations/collection/permissions.js
marla-singer: Done

organizations/logo/client/resumable.js
marla-singer: @brylie You are misread here. It shows message about user doesn't have permission to add logo.

brylie: Ah, OK. Disregard.

marla-singer:
@brylie I deleted the dublicated code and useless helper. The bug is resolved. There are no apparent reasons to not approve PR. Sometimes in the pursuit of "less code", "explicitly" and others "smart ideas from guides" we forgot about UX. We do the product for users as well, not only for developers.

organizations/client/profile/profile.html
brylie: Alright, we can make that improvement here. However, I still maintain that we should be checking whether the "user can view the organization" in the router, not the Organization Profile or its child templates.

marla-singer: All user can see all organization

organizations/client/profile/header/header.js
brylie: This check can be done via better subscription management, and a simple temlpate helper that fetches APIs.

1. toggle the subscription based on whether user can manage Organization
  - manager: subscribe to all Organization APIs
  - other users: subscribe to public APIs
2. In the template helper, check whether the `organization.currentUserCanManage` returns true
  - if true,  `apis = organizaiton.apis`
  - else, `apis = organization.publicApis`
3. `(if apis.length > 0) return apis`

marla-singer: I don't understand why your solution is better that my solution

brylie: In the above example, we have encapsulated the layers of abstraction into separate functions. At the outermost layer, the template helper, we have a function that checks the underlying layers. It should only be a few lines of code, and rather self-describing.

marla-singer: it is only your vision

organization_apis/collection/helpers/organizations.js
brylie: Right, which could be populated with APIs documents from any active subscription. In other words, it could return APIs that are not owned by the Organization.

It might just be better to have an `organization.publicApis` collection helper, and then use it elsewhere to return an array of Organization public APIs. Subsequent code can check the array length to create boolean logic.

marla-singer: I tested and it worked as expected

brylie: It works because you have correct subscriptions in place. This code **does not** check whether the APIs belong to the Organization, and so is misleading and potentially buggy.

brylie: Rather than doing `Apis.find().fetch()`, why not use the `organization.publicApis()` helper, and count the result length. That would be more safe.

apis/collection/server/publications.js
brylie: The branching logic is the "too many levels of abstraction". In other words, this function should only contain code to return a cursor of API documents, given an array of IDs, where the `api.isPublic === true`. 

The branching logic where we check for management permissions should be elsewhere, e.g. in a collection helper.

organization_apis/collection/helpers/organizations.js
marla-singer: Because we don't have `organization.publicApis()` helper. And if I create it, it will be duplicate of publication function `publicApis`.

apis/collection/server/publications.js
brylie: For the case where an user can see private APIs, use a different collection helper. E.g. `organization.apis()`.

Wrap the collection helpers in a conditional block that uses `organization.currentUserCanManage`, such as in the template helper described below.

marla-singer: I'm not going to redo that part, I don't have time for that

brylie: Likewise, the `organization.currentUserCanManage()` can be used inside of a composite publication, with a couple of secondary collection helpers that return cursors instead of arrays:

```js
// inside of composite publication
// at the child level where we retrieve organization APIs
function (organization) {
  let apisCursor;

  if (organization.currentUserCanManage()) {
    apisCursor = organization.allApisCursor();
  } else {
    apisCursor = organization.publicApisCursor();
  }

  return apisCursor;
}

Comment (brylie) [inline apis/collection/server/publications.js]: In the above examples, layers of abstraction are wrapped in their own function scope.

File: organizations/client/profile/apis/apis.js [hunk 15]
```diff
@@ -0,0 +1,100 @@
...
+    // Save selected option in reactive var
+    instance.selectedFilterOption.set(lifecycleParameter);
+
+    // Checking of filter was set
+    if (lifecycleParameter) {
+      // Filter data by selected parameter
+      managedApis = Apis.find({ lifecycleStatus: lifecycleParameter }).fetch();
+    } else {
+      // Otherwise show all managed apis
+      managedApis = Apis.find().fetch();
```
brylie: Please be more explicit with client side `find()` queries. Always pass in an explicit document ID or array of IDs when possible. Here, we can get an array of organization managed API IDs, using a collection helper. Then pass in the `apiIds` array in the query like `Apis.find({ _id: apiIds });`.

Alternatively, re-use the `organization.allApis` or `organization.allPublicApis` collection helper.

brylie: On second thought, this looks like a good use of a `templateInstance` reactive variable. There already seems to be an `instance.managedApis` reactive variable, so why not re-use it here? 

Move the logic that updates the `instance.managedApis` reactive variable into a template event handler.

organizations/client/profile/apis/apis.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    // Save list of managed APIs in instance reactive variable
+    instance.managedApis.set(managedApis);
```
brylie: Updating the reactive variable would be better done in a template instance event handler, rather than template helper.

organizations/client/profile/apis/apis.js
```diff
@@ -0,0 +1,100 @@
...
+
+    // Return count of managed/filtered apis
+    return managedApis.length;
+  },
+  apis () {
```
brylie: This template helper is short and easy to understand because it only `gets` a template instance reactive variable. The `apisCount` template helper should only be a couple of more lines in length than this helper.

organization_apis/collection/helpers/organizations.js
brylie: Sometimes publications and helpers duplicate a bit of logic, since they serve complementary purposes:

- publications send data to the client
- collection helpers rely on client/server data context to select specific documents or perform logic related to the collection

This code is confusing the distinction between collection helpers and publications.

brylie: Also of note, publications return database cursors (by definition), whereas collection helpers can return collection helpers, data objects/arrays, or any value. Publications are invoked on the server, when a subscription is called. Whereas collection helpers can run in the client or server, and rely on the data available in current scope.

organizations/client/profile/header/header.js
marla-singer: Deleted helper

marla-singer:
@brylie Current realization doesn't work. Publish function doesn't work via cursor. It throws error on server console:
**`Exception from sub userVisibleApis id LfXELt5vWghjNSwd6 Error: Meteor.userId can only be invoked in method calls. Use this.userId in publish functions`**
Link to this line https://github.com/Digipalvelutehdas/APIKA/blob/8b144b48fa9e535f0405a568bdbeb61b66ee3dc4/organization_apis/collection/helpers/organizations.js#L56

marla-singer:
@brylie I tried to provide user id from published function to helper as
**`organization.organization.userVisibleApisCursor(apiIds, this.userId);`** but the second parameter returns `undefined` in collection helper

brylie:
Dang it. I am sure there are practical reasons why we need to use `this.userId` inside of publications while `Meteor.userId` is used elsewhere. That is a good example of why consistency is important when developing a project or framework.

apis/collection/server/publications.js
```diff
@@ -21,23 +21,42 @@ Meteor.publish('apiBackend', (apiBackendId) => {
   return Apis.find({ _id: apiBackendId });
 });
 
-Meteor.publish('apisByIds', (apiIds) => {
+Meteor.publish('allOrganizationApisByIds', (apiIds) => {
```
brylie: Now that we have improved the overall code structure, we can update this publication.

1. Rename it to `allApisForSingleOrganization` (or something similar), and 
  - change the parameter to be `organizationId` or `organizationSlug`. 
2. Then, use the `organization.managedApiIds()` helper inside of the publication to get an array of `apiIds`.
3. Finally, use the `apiIds` array in the Apis collection query, as it exists now.

brylie: This improvement simplifies profile.js by moving the `managedApiIds` step into the publication.

brylie: Then, update profile.js so it reads like this:

```js
// Check Organization management privileges for current user
if (organization.currentUserCanManage()) {
  // User is allowed to see all public/private APIs 
  instance.subscribe(`allApisForSingleOrganization`, organizationSlug);
} else {
  // User is only allowed to see public APIs and those they manage
  instance.subscribe('allUserVisibleOrganizationApis', organizationSlug);
}
```

brylie: Basically, the subscriptions should mirror the collection helpers and template logic. These aren't arbitrary suggestions, but are intended to improve the code by using abstraction and fewer lines.

organization_apis/collection/helpers/organizations.js
```diff
@@ -36,20 +37,52 @@ Organizations.helpers({
...
 
-    // Find in managed APIs
-    const dbQuery = { _id: { $in: apiIds } };
+    // Make sure organization has APIs
+    if (apiIds.length > 0) {
+      // Placeholder for storage
+      let userId;
+
+      // Check if function parameter exists
+      if (currentUser) {
```
brylie: What happens if you check `if (this.userId) {}` here and set `userId = this.userId` if it is `true`?

brylie: It seems like that should work, since the previous problem was regarding `Meteor.userId()` being used inside of a publication. We are still using `Meteor.userId()` in this code, but won't use it if `this.userId` is defined. That avoids passing in the `currentUser` parameter.

organizations/client/profile/apis/show_apis/show_apis.js
```diff
@@ -0,0 +1,75 @@
...
+  const organization = instance.data.organization;
+
+  // Watching for changes of query parameters
+  instance.autorun(() => {
+    const managedApiIds = organization.managedApiIds();
+
+    // Placeholder
+    let managedApis;
+
+    // Get query parameter LifeCycle
```
brylie: Move these two lines closer to where the `lifecycleParameter` variable is used (e.g. line 26). That will make it easier to read what this function does, since the logic will jump around less.

brylie: Also, is it possible to have multiple instance.autorun() functions? If so, we might just separate the `lifecycleParameter` logic into its own autorun function, for clarity.

organizations/client/profile/apis/show_apis/show_apis.js
```diff
@@ -0,0 +1,75 @@
...
+      managedApis = Apis.find({ _id: { $in: managedApiIds } }).fetch();
+    }
+
+    // Save list of managed APIs in instance reactive variable
+    instance.managedApis.set(managedApis);
+  });
+});
+
+Template.organizationShowApis.helpers({
+  apisCount () {
```
brylie: Believe it or not, this helper is not necessary. Instead, do this inside of *show_apis.html*:

```html
<!-- line 7 -->
{{_ 'organizationProfile_smallText_apisCount' apis.length }}
```

Notice that I am using the `apis.length` property, since the template helper returns an array! :small_airplane:

brylie: The above simplification means that most of the code inside of `instance.autorun()` that deals with the Apis query can actually be moved directly inside of the `apis` template helper. 

The reason for this is that template helpers are reactive contexts, like autorun, so will automatically re-evaluate when any reactive data source within their function scope changes.

organizations/client/profile/profile.js
```diff
@@ -28,99 +19,35 @@ Template.organizationProfile.onCreated(function () {
...
-        const managedApiIds = organization.managedApiIds();
-        // Subscribe to Organization APIs documents
-        instance.subscribe('apisByIds', managedApiIds);
+        // If user is admin or organization manager
+        if (organization.currentUserCanManage()) {
+          // Get IDs of managed APIs via collection helper
+          const managedApiIds = organization.managedApiIds();
+
+          // Then publish all managed APIs of organization
+          instance.subscribe('allOrganizationApisByIds', managedApiIds);
```
brylie: Again, the logic getting an array of `managedApiIds` can be moved inside of the publication.

apis/collection/server/publications.js
```diff
@@ -19,23 +19,57 @@ Meteor.publish('apiBackend', (apiBackendId) => {
...
+        { isPublic: true },
+        { managerIds: userId },
+        { authorizedUserIds: userId },
+      ],
+    };
+  } else {
+    // Case: Anonymous users
+
+    // Show all public apis of organization
+    filteredApis = { _id: { $in: apiIds }, isPublic: true };
```
marla-singer: You looked an old code

organizations/client/profile/apis/show_apis/show_apis.js
marla-singer: As for me the template helper must be `clear` and returns the value.

marla-singer: Also we will call `apis()` twice in th template: `apis.length` and `apis=apis`.
In my case - On each render the *apis* value will be calculated only once in `autorun` 
In your case - On each render the *apis* value will be calculated twice

organizations/client/profile/apis/show_apis/show_apis.html
```diff
@@ -0,0 +1,21 @@
+<template name="organizationShowApis">
+  <div class="block-header">
+  <!-- Title -->
+  <h3 class="apis-block-title">
+    {{_ 'organizationProfile_bigText_apisCount'  }}
+    <small>
+      {{_ 'organizationProfile_smallText_apisCount' apisCount }}
```
brylie: OK, thiiiiiis line! There is a small improvement.

brylie: Whaaat happens if you do this:

```
{{_ 'organizationProfile_smallText_apisCount' apis.length }}
```

marla-singer: Agree :+1:

marla-singer:
@brylie Done

organizations/client/profile/apis/noApis.html
```diff
@@ -0,0 +1,10 @@
+<template name="organizationNoApis">
+  <div class="alert alert-warning" style="display: inline-block;">
+    <h4>
+      Orgazation doesn't have any connected APIs.
+    </h4>
+    {{# if organization.currentUserCanManage }}
+      You can connect one via button "Connect API to Organization"
```
marla-singer: It has done

organizations/client/profile/apis/noApis.html
```diff
@@ -0,0 +1,10 @@
+<template name="organizationNoApis">
+  <div class="alert alert-warning" style="display: inline-block;">
+    <h4>
+      Orgazation doesn't have any connected APIs.
```
marla-singer: It has done

PR #348, opened 2017-02-03
Bugfix/private api in profile with composite

brylie:


marla-singer:
Close it as no worked

